PR #33, opened 2020-11-24
Introduce shadowing

michaelvoronov:


stepper-lib/src/air/call/utils.rs
```diff
@@ -40,9 +40,34 @@ pub(super) fn set_local_call_result<'i>(
...
+                    // check that current execution flow is inside a fold block
+                    if exec_ctx.met_folds.is_empty() {
+                        // shadowing is allowed only inside fold blocks
+                        return Err(MultipleVariablesFound(entry.key().clone()));
+                    }
+
+                    match entry.get() {
+                        AValue::JValueRef(_) => {}
+                        // shadowing is allowed only for scalar values
+                        _ => return Err(MultipleVariablesFound(entry.key().clone())),
```
folex, line 56: I think this deserves a separate error with a special error message like `"ShadowError: Accumulator shadowing is not supported: {}", entry.key`

stepper-lib/src/air/call/utils.rs
```diff
@@ -40,9 +40,34 @@ pub(super) fn set_local_call_result<'i>(
...
+                    };
+
+                    entry.insert(AValue::JValueRef(result));
+                }
             };
+
+            if let Some(fold_block_name) = exec_ctx.met_folds.back() {
+                let fold_state = match exec_ctx.data_cache.get_mut(*fold_block_name) {
+                    Some(AValue::JValueFoldCursor(fold_state)) => fold_state,
+                    _ => unreachable!(),
```
folex, line 66: Maybe add a message, in case it's reached?

```suggestion
                    _ => unreachable!("fold block data must be represented as fold cursor"),
```

folex (approved):

PR #35, opened 2020-12-02
Add function to return AST as json

folex:


stepper/src/ast.rs
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+
+use stepper_lib::parse;
+
+/// Parse AIR script and return it as minified JSON
+pub fn ast(script: String) -> String {
+    let do_parse = || -> std::result::Result<_, Box<dyn std::error::Error>> {
+        let ast = parse(&script)?;
+        Ok(serde_json::to_string(&ast)?)
```
michaelvoronov, line 23: Is it possible to return just `serde_json::to_string(&ast)` here?

stepper-lib/src/execution/prolog.rs
```diff
@@ -77,3 +77,8 @@ pub(super) fn make_contexts(
 
     Ok((exec_ctx, call_evidence_ctx))
 }
+
+/// Parse an AIR script to AST
+pub fn parse(script: &str) -> Result<Instruction<'_>> {
+    Ok(*air_parser::parse(script).map_err(|msg| AquamarineError::AIRParseError(msg))?)
```
michaelvoronov, line 83: Suggest to create a new variable like this
```
let parsed_script = air_parser::parse(script).map_err(|msg| AquamarineError::AIRParseError(msg))?;
Ok(parsed_script)
```

michaelvoronov (approved):

PR #55, opened 2021-01-11
Refactoring and housekeeping

michaelvoronov:


stepper-lib/src/contexts/execution_context.rs
```diff
@@ -38,10 +39,10 @@ pub(crate) struct ExecutionCtx<'i> {
 
     /// Indicates that previous executed subtree is complete.
     /// A subtree treats as a complete if all subtree elements satisfy the following rules:
-    ///   - at least one of par subtrees is complete
-    ///   - non-thrown subtree of xor is complete
-    ///   - all of seq subtrees are complete
-    ///   - call executes successfully (call evidence equals to Executed)
+    ///   - at least one of par subtrees is completed
+    ///   - non-thrown subtree of xor is completed
```
folex, line 43: Do I understand right that this line means:

> at least one of xor substree is completed without an error

?

or does it mean that the first subscribe as in `(xor first second)` is completed?

Also, what would happen if this condition isn't met? Will script hang, or is it guaranteed to return with an error, thus kind of completing the xor?

stepper-lib/src/contexts/execution_trace_context/executed_state.rs
```diff
@@ -0,0 +1,55 @@
...
+
+use serde::Deserialize;
+use serde::Serialize;
+use std::rc::Rc;
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum CallResult {
+    /// Request was sent to a target node by node with such public key and it shouldn't be called again.
+    RequestSentBy(String),
```
folex, line 27: 🥇

stepper-lib/src/log_targets.rs
```diff
@@ -26,34 +26,34 @@ pub const NEXT_PEER_PKS: &str = "next_peer_pks";
...
-pub const EVIDENCE_PATH_MERGE: &str = "evidence_merge";
+/// Print out logs at the executed states merging stage.
+pub const EXECUTED_TRACE_MERGE: &str = "executed_trace_merge";
 
 /// Print out running arguments and params of a script.
 pub const RUN_PARAMS: &str = "initial_params";
 
 /// Print out state of data cache at the beginning of each instruction execution.
-pub const EVIDENCE_CHANGING: &str = "evidence_changing";
+pub const EXECUTED_STATE_CHANGING: &str = "executed_trace_changing";
```
folex, line 45: I think it should be `EXECUTION_STATE_CHANGE = "execution_state_change"`

stepper-lib/src/execution/air/mod.rs
```diff
@@ -80,19 +75,19 @@ macro_rules! log_instruction {
         );
 
         log::debug!(
-            target: crate::log_targets::CALL_EVIDENCE_PATH,
-            "  current call evidence path: {:?}",
-            $call_ctx.current_path
+            target: crate::log_targets::EXECUTED_TRACE,
```
folex, line 78: Should it be EXECUTION_TRACE or EXECUTED_TRACE? What's the intention:
- trace of what? => of execution
- what kind of trace? => executed trace (and there might be an execut**ING** trace or something)

folex (approved):

PR #57, opened 2021-01-15
Make xor catch more execution errors; fix joinable behaviour

michaelvoronov:


stepper-lib/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "stepper-lib"
-version = "0.2.2"
+version = "0.2.3"
```
folex, line 3: I think this deserves a minor version increment to express major change in behavior

folex (approved):

stepper-lib/Cargo.toml
michaelvoronov: Sounds fair

PR #60, opened 2021-01-24
Introduce the match instruction

michaelvoronov:


folex (approved):
Maybe remove `Value` from `CallArgValue`, `MatchableValue`, etc?

michaelvoronov:
We discussed that in PM and decided not to do it now and leave for further boxed value integration due to a name clashing.

PR #64, opened 2021-02-11
Support numbers and booleans

michaelvoronov:


crates/air-parser/src/parser/lexer/call_variable_parser.rs
```diff
@@ -0,0 +1,242 @@
...
+
+    fn try_parse_as_number(&mut self) -> LexerResult<()> {
+        if self.try_parse_sign() || self.try_parse_digit() || self.try_parse_float_dot() {
+            return Ok(());
+        }
+
+        self.handle_non_digit()
+    }
+
+    fn try_parse_sign(&self) -> bool {
```
folex, line 107: If a function does unwrap, maybe call it `unchecked` instead of `try`? I propose the following scheme:
1. `try_xx` return LexerResult, unwraps are prohibited
2. `unchecked_xx` or `xx_unchecked` do unwrap

crates/air-parser/src/parser/lexer/call_variable_parser.rs
```diff
@@ -0,0 +1,242 @@
+/*
```
folex, line 1: I love the code here! So clean and tidy, awesome work!

crates/air-parser/src/parser/lexer/call_variable_parser.rs
```diff
@@ -0,0 +1,242 @@
...
+
+    fn try_parse_float_dot(&mut self) -> bool {
+        self.try_parse_first_met_dot()
+    }
+
+    fn handle_non_digit(&mut self) -> LexerResult<()> {
+        self.state.non_numeric_met = true;
+
+        if self.state.is_sign_met {
+            return Err(LexerError::UnallowedCharInNumber(
```
folex, line 124: I don't understand why this error is thrown. You're handling non-digit (`handle_non_digit `), and that means that you're expecting the current context not to be a number, right?  And yet you're throwing `UnallowedCharIn*Number*`. That seems contradictory to me, I think I'm missing something.

folex: What will happen on input like `1a`? `is_sign_met` will be false, so it seems like you would allow `1`, and then try to parse `a` as a variable. Is that so? 

Maybe you need something like `self.numeric_started`?

crates/air-parser/src/parser/lexer/call_variable_parser.rs
```diff
@@ -0,0 +1,242 @@
...
+
+    fn aqua_alphanumeric(&self) -> bool {
+        super::is_aqua_alphanumeric(self.state.current_char.unwrap())
+    }
+
+    fn json_path_allowed_char(&self) -> bool {
+        super::is_json_path_allowed_char(self.state.current_char.unwrap())
+    }
+
+    fn current_pos(&self) -> usize {
```
folex, line 196: `current_pos_unchecked`?

crates/air-parser/src/parser/lexer/errors.rs
```diff
@@ -32,6 +35,20 @@ pub enum LexerError {
 
     #[error("invalid character in json path")]
     InvalidJsonPath(usize, usize),
+
+    #[error("this signed digit could have only digits or one dot")]
+    UnallowedCharInNumber(usize, usize),
+
+    #[error("{2}")]
```
folex, line 42: could 
```
    #[error(transparent)]
    ParseIntError(usize, usize, #[from] anyhow::Error), 
```
be used here?

crates/air-parser/src/parser/lexer/errors.rs
```diff
@@ -32,6 +35,20 @@ pub enum LexerError {
...
+    ParseIntError(usize, usize, ParseIntError),
+
+    #[error("{2}")]
+    ParseFloatError(usize, usize, ParseFloatError),
```
folex, line 46: use `#[source]` for error chaining:

```suggestion
    ParseFloatError(usize, usize,  #[source] ParseFloatError),
```

crates/air-parser/src/parser/lexer/tests.rs
```diff
@@ -172,6 +173,100 @@ fn string_literal() {
     );
 }
```
folex, line 175: need a test for `1abc` to fail

crates/air-parser/src/parser/lexer/tests.rs
```diff
@@ -184,6 +279,18 @@ fn json_path() {
...
 
+#[test]
+fn json_path_numbers() {
+    // this json path contains all allowed in json path charactes
+    const JSON_PATH: &str = r#"12345.$[$@[]():?.*,"!]"#;
```
folex, line 285: why allow json path after a number?

folex (approved):

crates/air-parser/src/parser/lexer/call_variable_parser.rs
```diff
@@ -0,0 +1,242 @@
...
+                self.try_parse_as_variable()?;
+            }
+
+            self.state.is_first_char = false;
+        }
+
+        self.to_token()
+    }
+
+    fn next_char(&mut self) -> bool {
```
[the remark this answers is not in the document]
michaelvoronov: it's impossible because `try_parse_as_number` checks `is_first_char` state after the first `next_chat` call.

crates/air-parser/src/parser/lexer/call_variable_parser.rs
```diff
@@ -0,0 +1,242 @@
...
+
+    pub(self) fn try_parse(mut self) -> LexerResult<Token<'input>> {
+        while self.next_char() {
+            if self.is_it_possible_to_parse_as_number() {
+                self.try_parse_as_number()?;
+            } else {
+                self.try_parse_as_variable()?;
+            }
+
+            self.state.is_first_char = false;
```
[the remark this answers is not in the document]
michaelvoronov: ditto

crates/air-parser/src/parser/lexer/call_variable_parser.rs
michaelvoronov: We discussed that and decided that for now it's ok if variable starts with a digit.

crates/air-parser/src/parser/lexer/errors.rs
michaelvoronov: Afaik `#[error(transparent)]` requires exactly one field in a enum variant.

crates/air-parser/src/parser/lexer/errors.rs
michaelvoronov: Are there any benefits of using `#[source]` inside a enum?

PR #65, opened 2021-02-12
Switch to cjs

coder11:
Switch to CommonJS modules instead of ESM

### DIFF
--- a/.github/workflows/publish_interpreter.yml
+++ b/.github/workflows/publish_interpreter.yml
@@ -102,23 +102,28 @@ jobs:
       - name: Generate aquamarine.wasm.base64.js
         run: |
           cat << EOF > pkg/aquamarine.wasm.base64.js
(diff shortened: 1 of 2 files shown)


.github/workflows/publish_interpreter.yml
```diff
@@ -102,23 +102,28 @@ jobs:
       - name: Generate aquamarine.wasm.base64.js
         run: |
           cat << EOF > pkg/aquamarine.wasm.base64.js
-          export const wasmBs64 = "$(base64 -w0 pkg/aquamarine_client_bg.wasm)";
+          module.exports = "$(base64 -w0 pkg/aquamarine_client_bg.wasm)";
```
DieMyst, line 105: if you know how it possible to use raw bytes instead of base64 string - it will be great

DieMyst (approved):

.github/workflows/publish_interpreter.yml
coder11: I have some ideas how to improve it. Will do it a little bit later

PR #69, opened 2021-02-18
Improve flattening

michaelvoronov:


interpreter-lib/src/execution/errors.rs
```diff
@@ -91,6 +91,10 @@ pub(crate) enum ExecutionError {
     /// This error type is produced by a mismatch to notify xor that compared values aren't equal.
     #[error("mismatch is used without corresponding xor")]
     MismatchWithoutXorError,
+
+    /// This error type is produced by a mismatch to notify xor that compared values aren't equal.
+    #[error("jvalue '{0}' can't be flattened, to be flattened a jvalue should has an array type and consists only one value")]
```
folex, line 96:
```suggestion
    #[error("jvalue '{0}' can't be flattened, to be flattened a jvalue should have an array type and consists only one value")]
```

interpreter/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "aquamarine"
-version = "0.6.0"
+version = "0.6.1"
```
folex, line 3: I think these are breaking changes, so maybe `0.7.0`?

folex (approved):

PR #70, opened 2021-02-25
Add a new test for join behaviour

michaelvoronov:


interpreter/src/logger.rs
```diff
@@ -18,7 +18,7 @@ use interpreter_lib::log_targets::TARGET_MAP;
 
 use log::Level as LogLevel;
 
-pub const DEFAULT_LOG_LEVEL: LogLevel = LogLevel::Info;
+pub const DEFAULT_LOG_LEVEL: LogLevel = LogLevel::Trace;
```
folex, line 21: Should it be reverted to INFO?

folex (approved):
Let's lift `join` log to `warn` level until we have `#`

PR #72, opened 2021-02-28
Fix logger

michaelvoronov:


interpreter-lib/tests/security_tetraplets.rs
```diff
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+/*
```
folex, line 17: is it ok to keep that commented out?

interpreter/src/fce.rs
```diff
@@ -17,7 +17,7 @@
 #![allow(improper_ctypes)]
 #![warn(rust_2018_idioms)]
 #![deny(
-    dead_code,
+    //dead_code,
```
folex, line 20: ditto

folex (approved):

PR #76, opened 2021-03-17
Basic validator for variables

michaelvoronov:


crates/air-parser/src/parser/validator.rs
```diff
@@ -0,0 +1,209 @@
...
+        }
+    }
+
+    fn meet_variable(&mut self, name: &'i str, span: Span) {
+        if !contains(&self.met_variables, name, span) && !contains(&self.met_iterable, name, span) {
+            self.unresolved_variables.insert(name, span);
+        }
+    }
+
+    fn meet_call_output(&mut self, call_output: &CallOutputValue<'i>, span: Span) {
```
folex: Let's rename it to `meet_variable_definition` or `define_variable`?

crates/air-parser/src/parser/validator.rs
```diff
@@ -0,0 +1,209 @@
...
+    }
+}
+
+fn contains(multimap: &MultiMap<&str, Span>, key: &str, key_span: Span) -> bool {
+    let found_spans = match multimap.get_vec(key) {
+        Some(found_spans) => found_spans,
+        None => return false,
+    };
+
+    found_spans.iter().any(|s| s < &key_span)
```
folex, line 187: Maybe only store a minimal found span? I think this would allow to avoid `multimap` using `hashmap` instead, and also would make `contains` a constant time operation.

I mean, on `insert` it is possible to compare new span with existing, and choose the minimum of the two

interpreter-lib/src/execution/air/fold.rs
```diff
@@ -337,7 +337,7 @@ mod tests {
         );
 
         let res = call_vm!(set_variable_vm, "", empty_fold.clone(), "[]", "[]");
-        let res = call_vm!(vm, "", empty_fold, "[]", res.data);
+        //let res = call_vm!(vm, "", empty_fold, "[]", res.data);
```
folex, line 340: intended?

interpreter-lib/src/execution/air/fold.rs
```diff
@@ -320,7 +320,7 @@ mod tests {
         use crate::contexts::execution_trace::CallResult::*;
         use crate::contexts::execution_trace::ExecutedState::*;
 
-        let mut vm = create_aqua_vm(echo_number_call_service(), "A");
+        let mut _vm = create_aqua_vm(echo_number_call_service(), "A");
```
folex, line 323: intended?

folex (commented):
- I think it would be beneficial to document meaning of fields for `ast::Fold` because `iterator` and `iterable` are so alike and it's hard to tell one from the other sometimes

- I'm not sure, but maybe functions `meet_something` should be actually `met_something`? Like, to designate what happened vs. imperative/infinitive form

crates/air-parser/src/parser/validator.rs
```diff
@@ -0,0 +1,209 @@
...
+    fn meet_iterable_value(&mut self, iterable_value: &IterableValue<'i>, span: Span) {
+        match iterable_value {
+            IterableValue::JsonPath { variable, .. } => self.meet_variable(variable, span),
+            IterableValue::Variable(variable) => self.meet_variable(variable, span),
+        }
+    }
+
+    fn meet_iterator(&mut self, iterator: &'i str, span: Span) {
+        self.met_iterable.insert(iterator, span);
+    }
```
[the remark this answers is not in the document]
michaelvoronov: `meet_iterable` considers iterables, values that should be defined on some call. And `meet_iterators` - iterators, that defined in `fold`.

I renamed `self.met_iterable` to `self.met_iterators`

folex (approved):

PR #77, opened 2021-03-23
Make match mismatch joinable

michaelvoronov:


crates/air-parser/src/parser/validator.rs
```diff
@@ -175,6 +185,15 @@ impl<'i> VariableValidator<'i> {
         }
     }
 
+    fn met_matchable(&mut self, matchable: &MatchableValue<'i>, span: Span) {
+        match matchable {
+            MatchableValue::Number(_) | MatchableValue::Boolean(_) | MatchableValue::Literal(_) => {
+            }
```
folex, line 191:
```suggestion
            MatchableValue::Number(_) | MatchableValue::Boolean(_) | MatchableValue::Literal(_) => {}
```

interpreter-lib/src/execution/air/mod.rs
```diff
@@ -147,3 +147,59 @@ macro_rules! log_instruction {
...
     };
 }
+
+/// This macro converts joinable errors to Ok and sets subtree complete to true.
+#[macro_export]
+macro_rules! joinable_complete {
+    ($cmd:expr, $exec_ctx:expr) => {
+        match $cmd {
+            Err(e) if crate::execution::air::is_joinable_error_type(&e) => {
+                $exec_ctx.subtree_complete = false;
```
folex, line 157: comment says
> sets subtree complete to true.

code does
`$exec_ctx.subtree_complete = false;`

true or false, which one is right?

folex (changes requested):

interpreter-lib/src/execution/air/mod.rs
michaelvoronov: The comment was incorrect.

folex (approved):

PR #100, opened 2021-05-10
Checking for undefined textencoder

coder11:
Fixing `ReferenceError: TextEncoder is not defined` error


DieMyst (approved):

npm/package.json
```diff
@@ -20,6 +20,7 @@
     },
     "devDependencies": {
         "@types/jest": "^26.0.23",
+        "@types/node": "^15.0.2",
```
folex, line 23: Let's use 14 if possible

npm/src/wrapper.ts
```diff
@@ -97,9 +99,9 @@ function getInt32Memory0(wasm) {
     return cachegetInt32Memory0;
 }
 
-let cachedTextDecoder = new TextDecoder('utf-8', { ignoreBOM: true, fatal: true });
+const lTextDecoder = typeof TextDecoder === 'undefined' ? module.require('util').TextDecoder : TextDecoder;
 
-cachedTextDecoder.decode();
```
folex, line 102: is it OK that you don't call `decode` anymore?

coder11: Well, seems like this line was added by accident

npm/package.json
coder11: ok

mikevoronov (approved):

folex (approved):

PR #102, opened 2021-05-12
Move to sdk 0.6.5

mikevoronov:


air-interpreter/Cargo.toml
```diff
@@ -26,4 +26,4 @@ serde = { version = "=1.0.118", features = [ "derive", "rc" ] }
 serde_json = "1.0"
```
boneyard93501, line 26: 1.0.56 ?

air-interpreter/Cargo.toml
```diff
@@ -11,12 +11,12 @@ path = "src/wasm_bindgen.rs"
...
-path = "src/fce.rs"
+path = "src/marine.rs"
 
 [dependencies]
 air = { path = "../air" }
 
-fluence = { version = "0.6.2", features = ["logger"] }
+fluence = { version = "0.6.5", features = ["logger"] }
 
 # Keep 0.2.65 until this is resolved https://github.com/rustwasm/wasm-pack/issues/886
```
boneyard93501, line 21: isn't https://github.com/rustwasm/wasm-pack/issues/886 resolved ?

mikevoronov: It is, but it has to be checked and I think it deserves a separate PR.
@coder11 could you please a look?

air-interpreter/Cargo.toml
mikevoronov: We could fix it, yep, but is it worth so?

boneyard93501: probably not

air-interpreter/Cargo.toml
boneyard93501: yep

air-interpreter/Cargo.toml
folex: It uses 1.0.61 judging by Cargo.lock

folex (approved):

PR #112, opened 2021-06-02
Make %last_error% object

mikevoronov:


air/src/execution/utils/resolve.rs
```diff
@@ -53,27 +55,18 @@ fn prepare_consts(arg: impl Into<JValue>, ctx: &ExecutionCtx<'_>) -> ExecutionRe
...
-        }
+fn prepare_last_error(
+    path: &LastErrorPath,
+    ctx: &ExecutionCtx<'_>,
+) -> ExecutionResult<(JValue, Vec<SecurityTetraplet>)> {
+    let LastErrorWithTetraplets { last_error, tetraplets } = ctx.last_error();
+    let jvalue = match path {
+        LastErrorPath::Instruction => JValue::String(last_error.instruction),
+        LastErrorPath::Message => JValue::String(last_error.msg),
+        LastErrorPath::None => serde_json::to_value(last_error).expect("the default serializer shouldn't fail"),
```
folex, line 66: you can do `json!(last_error)`. It panics inside, but looks nicer :D

crates/air-parser/src/parser/lexer/air_lexer.rs
```diff
@@ -194,6 +195,25 @@ fn string_to_token(input: &str, start_pos: usize) -> LexerResult<Token> {
     }
 }
 
+fn parse_last_error(input: &str, start_pos: usize) -> LexerResult<Token<'_>> {
+    let last_error_size = LAST_ERROR.len();
+    let last_error_path = match &input[last_error_size..] {
```
folex, line 200: Maybe also account for possible `$`? Like `%last_error%.$.instruction`

folex (approved):

air/src/execution/utils/resolve.rs
mikevoronov: Yep, will do, it really looks nicer.

crates/air-parser/src/parser/lexer/air_lexer.rs
mikevoronov: Hmm, I am not sure, we're going to get rid of json path in the nearest future.

mikevoronov: Discuss it in PM and decided to use `.$` here for consistency.

PR #116, opened 2021-06-04
Check that json path is allowed

mikevoronov:


folex (approved):
Maybe add a test that it now throws an error on String?

mikevoronov:
> Maybe add a test that it now throws an error on String?

Done

PR #130, opened 2021-08-21
Make interpreter async

mikevoronov:
Resolves https://github.com/fluencelabs/aquavm/issues/140.


avm/server/src/lib.rs
```diff
@@ -46,7 +45,13 @@ pub use fluence_faas::HostImportError;
 pub use fluence_faas::IType;
 pub use fluence_faas::IValue;
 
+pub use air_interpreter_interface::CallRequests;
+pub use air_interpreter_interface::CallResults;
 pub use air_interpreter_interface::InterpreterOutcome;
```
folex, line 50: Is there any reason to keep InterpreterOutcome in exports of avm-server? I'd think AVMOutcome supersedes it

crates/interpreter-interface/src/call_request_parameters.rs
```diff
@@ -0,0 +1,47 @@
...
+
+pub type CallRequests = HashMap<u32, CallRequestParams>;
+
+#[marine]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct CallRequestParams {
+    pub service_id: String,
+    pub function_name: String,
+    pub arguments: String,
+    pub tetraplets: String,
```
folex, line 30: Let's add some docs here pls

avm/server/src/lib.rs
mikevoronov: Nope, you're right.

crates/interpreter-interface/src/call_request_parameters.rs
```diff
@@ -0,0 +1,56 @@
...
+#[marine]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct CallRequestParams {
+    /// Id of a service that should be called.
+    pub service_id: String,
+
+    /// Name of a function from service identified by service_id that should be called.
+    pub function_name: String,
+
+    /// Serialized to JSON string Vec<IValue> of arguments that should be passed to a service.
```
folex, line 35: It's a serialized Vec<JValue>, right? `IValue` is a typo

mikevoronov: Yes, you are right, thanks!

coder11 (approved):

PR #152, opened 2021-10-05
Decouple log targets to a separate crate

mikevoronov:


air/src/runner.rs
```diff
@@ -16,16 +16,17 @@
...
 use crate::execution_step::ExecutionError;
-use crate::execution_step::{Catchable, TraceHandler};
-use crate::log_targets::RUN_PARAMS;
+use crate::execution_step::TraceHandler;
 use crate::preparation_step::prepare;
 use crate::preparation_step::PreparationDescriptor;
 
 use air_interpreter_interface::InterpreterOutcome;
 use air_interpreter_interface::RunParameters;
+use air_log_targets::RUN_PARAMS;
```
ValeryAntopol, line 29: I suggest new line after this line

crates/air-lib/trace-handler/Cargo.toml
```diff
@@ -15,8 +15,9 @@ name = "air_trace_handler"
 path = "src/lib.rs"
 
 [dependencies]
-air-interpreter-data = { path = "../interpreter-data" }
 air-parser = { path = "../air-parser" }
+air-interpreter-data = { path = "../interpreter-data" }
+air-log-targets = { path = "../log-targets" }
```
ValeryAntopol, line 20: If we need lexicographic ordering, then air-parser must be the last

ValeryAntopol (approved):

PR #164, opened 2021-11-02
Deduplicate SecuityTetraplet with marine-rs-sdk

mikevoronov:


crates/air-lib/polyplets/src/lib.rs
```diff
@@ -25,6 +25,16 @@
     unreachable_patterns
 )]
 
+extern crate self as marine_rs_sdk;
```
folex, line 28: Wow, why is that needed?

ValeryAntopol: #[marine] generates code that uses marine_rs_sdk::internal, so it is needed to provide it without depending on marine_rs_sdk

folex: I see. Clever trick! I think it deserves a comment.

crates/air-lib/polyplets/src/triplet.rs
```diff
@@ -19,6 +19,8 @@ use serde::Serialize;
 
 /// ResolvedTriplet represents peer network location with all
 /// variables, literals and etc resolved into final string.
+/// This structure contains a subset of value that
+/// SecurityTetraplet consist of
```
ValeryAntopol, line 23: value -> values
consist -> consists

ValeryAntopol (approved):

PR #166, opened 2021-11-04
fix behaviour with compare values after applying lambda

### LINKED ISSUE #165: Bug: matching values after applying lambda is broken
This bug was introduced after switching to lamda from json path, the crux of it is that a comparator additionaly wraps values obtained from lambda in `JValue::Array` and doesn't consider flattening.

So in this example
```
(xor
  (match result.$.success true
    (ap 1 value)
  )
  (ap 2 value)
)

(call %init_peer_id% ("callbackSrv" "response") [scalar])
```
`value` will be `2` although `result.$.success` returns `true`.

mikevoronov:
Resolves #165

### DIFF
--- a/air/src/execution_step/air/compare_matchable/comparator.rs
+++ b/air/src/execution_step/air/compare_matchable/comparator.rs
@@ -115,9 +115,9 @@ fn compare_matchable<'ctx>(
             let jvaluable = resolve_ast_variable(&vl.variable, exec_ctx)?;
             let jvalues = jvaluable.apply_lambda(&vl.lambda)?;
 
-            // it's known that apply_lambda always returns array with one value that is intended
-            // to support multi-return in the future, this check is needed just in case and
-            // should be refactored after introducing boxed values
+            // TODO: it's known that apply_lambda always returns array with one value that is
+            // intended to support multi-return in the future, this check is needed just in
+            // case and should be refactored after introducing boxed values
             if jvalues.len() != 1 {
                 return Ok(false);
             }
--- a/air/src/execution_step/air/compare_matchable/comparator.rs
+++ b/air/src/execution_step/air/compare_matchable/comparator.rs
@@ -115,10 +115,14 @@ fn compare_matchable<'ctx>(
             let jvaluable = resolve_ast_variable(&vl.variable, exec_ctx)?;
             let jvalues = jvaluable.apply_lambda(&vl.lambda)?;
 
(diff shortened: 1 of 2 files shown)


air/src/execution_step/air/compare_matchable/comparator.rs
```diff
@@ -115,10 +115,14 @@ fn compare_matchable<'ctx>(
             let jvaluable = resolve_ast_variable(&vl.variable, exec_ctx)?;
             let jvalues = jvaluable.apply_lambda(&vl.lambda)?;
 
-            let jvalue = jvalues.into_iter().cloned().collect::<Vec<_>>();
-            let jvalue = JValue::Array(jvalue);
-
-            Ok(comparator(Cow::Owned(jvalue)))
+            // it's known that apply_lambda always returns array with one value that is intended
+            // to support multi-return in the future, this check is needed just in case and
+            // should be refactored after introducing boxed values
```
folex, line 120: Let's mark it as `TODO`

air/tests/test_module/instructions/match_.rs
```diff
@@ -234,3 +234,39 @@ fn match_with_two_xors() {
...
+    let echo_peer_id = "echo_peer_id";
+    let mut echo_peer = create_avm(echo_call_service(), echo_peer_id);
+
+    let script = format!(
+        r#"
+        (seq
+            (call "{0}" ("" "") ["set_result"] result)
+            (seq
+                (xor
+                    (match result.$.success! true
```
folex, line 256: Does `!` matter here?

folex (approved):

air/tests/test_module/instructions/match_.rs
folex: If not, maybe modify the test so it tests both `$.success!` and `$.success`?

folex: I'd be sleeping better if I know that this behavior is captured in tests :)

mikevoronov: No, `!` is ignored by lambda, it's possible to write it for backward compatibility, but it doesn't really make sense.

folex: Ah, so it's ignored everywhere? That's great :) No need for tests modification then, agree.

mikevoronov: Yep, it ignored everywhere

PR #170, opened 2021-11-10
Change preparation error mapping

mikevoronov:
The interpreter returns 0 if supplied air is invalid due to the automatic error to int mapping by `strum`. This PR changes this scheme and force the interpreter to return not 0 in case of any error.

### DIFF
--- a/air/src/runner/outcome.rs
+++ b/air/src/runner/outcome.rs
@@ -31,9 +31,6 @@ use std::collections::HashMap;
 use std::hash::Hash;
 use std::rc::Rc;
 
-const PREPARATION_ERRORS_START_ID: i32 = 1;
-const EXECUTION_ERRORS_START_ID: i32 = 1000;
-
 /// Create InterpreterOutcome from supplied execution context and trace handler,
 /// set ret_code to INTERPRETER_SUCCESS.
 pub(crate) fn from_success_result(exec_ctx: ExecutionCtx<'_>, trace_handler: TraceHandler) -> InterpreterOutcome {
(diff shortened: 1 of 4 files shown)


coder11 (approved):

air/src/runner/outcome.rs
```diff
@@ -59,6 +60,7 @@ pub(crate) fn from_success_result(exec_ctx: ExecutionCtx<'_>, trace_handler: Tra
 /// set ret_code based on the error.
 pub(crate) fn from_preparation_error(data: impl Into<Vec<u8>>, err: PreparationError) -> InterpreterOutcome {
     let ret_code = err.to_error_code() as i32;
+    let ret_code = PREPARATION_ERRORS_START_ID + ret_code;
```
folex, line 63: Maybe move that into `to_error_code` implementation?

Regarding `PREPARATION_ERRORS_START_ID`, maybe it's not needed? This seems like a very general approach: all error codes must start with `1`, not just `PreparationError`. And I don't think `1` will be changed to anything else, will it?

folex: I think it's possible to implement `to_error_code` for [IntoEnumIterator](https://docs.rs/strum/0.22.0/strum/trait.IntoEnumIterator.html) trait. For example, `ExecutionError` also uses same approach in `to_error_code`, it would makes sense to start from `1` there, too.

mikevoronov: ExecutionErrors starts from `EXECUTION_ERRORS_START_ID = 100`, it's intended to distinguish between them

mikevoronov: regarding moving that into `to_error_code` - it makes sense, thanks

PR #175, opened 2021-11-22
Add test for par ap behaviour

mikevoronov:


air/tests/test_module/instructions/ap.rs
```diff
@@ -185,3 +187,32 @@ fn ap_with_dst_stream() {
...
+        (par
+            (call "{2}" ("peer" "timeout") [] join_it)
+            (seq
+                (par
+                    (call "{0}" ("peer" "timeout") [join_it] $result)
+                    (ap "fast_result" $result)
+                )
+                (call "{1}" ("op" "return") [$result.$[0]])
+            )
+        )
```
folex, line 209: - Since `vm_1` is the one who initiates the script, `op return` should be executed there. 
- `peer timeout` should be executed on vm_2 for this test to keep original meaning
-  There's no vm_3 in the original test, but I think that doesn't matter much

```suggestion
        (par
            (call "{2}" ("peer" "timeout") [] join_it)
            (seq
                (par
                    (call "{1}" ("peer" "timeout") [join_it] $result)
                    (ap "fast_result" $result)
                )
                (call "{0}" ("op" "return") [$result.$[0]])
            )
        )
```

air/tests/test_module/instructions/ap.rs
```diff
@@ -185,3 +187,32 @@ fn ap_with_dst_stream() {
...
+        "#,
+        vm_1_peer_id, vm_2_peer_id, vm_3_peer_id
+    );
+
+    let mut result = checked_call_vm!(vm_1, "", script, "", "");
+    let actual_next_peers: HashSet<_> = result.next_peer_pks.drain(..).collect();
+    let expected_next_peers: HashSet<_> = maplit::hashset!(vm_2_peer_id.to_string(), vm_3_peer_id.to_string());
+    assert_eq!(actual_next_peers, expected_next_peers);
```
folex, line 217: This isn't a complete test of this script. It only checks the first execution, but there's more. After checking next_peer_pks, you need to continue executing the particle. 

1. from vm_1 it hops to vm_2, vm_3
2. on vm_3, it sets `join_it` and hops to vm_2
3. on vm_2, it waits for `join_it`
4. on vm_2, it writes to `$result` and hops to vm_1
5. vm_1 executes `op return`

Originally, the problem happens on step 4: it doesn't hop `vm_2 => vm_1`

folex (commented):
Overall, I find it really hard to follow `vm_1, vm_2, vm_3` naming. I suggest to use the following (in pseudo-code):
```
let client = vm_1;
let relay = vm_2;
let other = vm_3;

let script = format!(
    r#"
    (par
        (call "{other}" ("peer" "timeout") [] join_it)
        (seq
            (par
                (call "{client}" ("peer" "timeout") [join_it] $result)
                (ap "fast_result" $result)
            )
            (call "{relay}" ("op" "return") [$result.$[0]])
        )
    )
    "#
);

client.execute(script);

let expected = vec![relay, other];
```

`f!` macro from https://lib.rs/fstrings will allow you to use `{client}` instead of `{0}` thus making the test more readable.

folex (approved):

mikevoronov:
Closes #178

PR #182, opened 2021-11-26
Track previous call state after join behaviour

### LINKED ISSUE #179: Bug: executed state doesn't preserved on join behaviour
The interpreter doesn't preserve `sent_by` state on join behaviour.

mikevoronov:
Resolves #179.

### DIFF
--- a/air/src/execution_step/air/call/prev_result_handler.rs
+++ b/air/src/execution_step/air/call/prev_result_handler.rs
@@ -26,8 +26,8 @@ use air_parser::ast::CallOutputValue;
 use air_trace_handler::TraceHandler;
 
 pub(crate) struct StateDescriptor {
-    pub(crate) should_execute: bool,
-    pub(crate) prev_state: Option<CallResult>,
+    should_execute: bool,
+    prev_state: Option<CallResult>,
 }
 
 /// This function looks at the existing call state, validates it,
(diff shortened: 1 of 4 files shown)


air/tests/test_module/instructions/fold.rs
```diff
@@ -234,8 +234,9 @@ fn fold_with_join() {
     let result = checked_call_vm!(set_variable_vm, "", fold_with_join, "", "");
     let result = checked_call_vm!(vm, "", fold_with_join, "", result.data);
 
+    print_trace(&result, "result");
```
ValeryAntopol, line 237: Forgotten debug output(maybe its ok in tests)

air/src/execution_step/air/call/prev_result_handler.rs
```diff
@@ -151,3 +156,50 @@ fn try_to_service_result(
...
+
+impl StateDescriptor {
+    pub(crate) fn executed() -> Self {
+        Self {
+            should_execute: false,
+            prev_state: None,
+        }
+    }
+
+    pub(crate) fn not_prepared(prev_state: CallResult) -> Self {
```
ValeryAntopol, line 168: Maybe `not_ready` is better, but not sure

ValeryAntopol (approved):

PR #194, opened 2021-12-15
Add variables names in resolve errors

### LINKED ISSUE #171: Write variables names in errors if needed
it is not clear where the error occurred. In `-relay-`, or in `[]`, or in `"op"`?
```
instruction: 'call -relay- ("op" "noop") [] ',
msg: "expected JValue type 'string', but got 'null' JValue"
```

It will be better if the variable name will be written, like `" In '-relay-" expected JValue type 'string', but got 'null' JValue"`
Or `got 'null' in peerId target`.

mikevoronov:
Resolves #171.


air/tests/test_module/integration/last_error.rs
```diff
@@ -18,6 +18,9 @@ use air::LastError;
 use air::SecurityTetraplet;
 use air_test_utils::prelude::*;
 
+use fstrings::f;
+use fstrings::format_args_f;
```
ValeryAntopol, line 22: I can't find any place where it is used, maybe we don't need this import?

air/src/execution_step/errors.rs
```diff
@@ -59,8 +59,12 @@ pub(crate) enum ExecutionError {
     EmptyStreamLambdaError,
 
     /// Provided JValue has incompatible type with a requested one.
-    #[error("expected JValue type '{1}', but got '{0}' JValue")]
-    IncompatibleJValueType(JValue, &'static str),
+    #[error("expected JValue type '{expected_value_type}' for variable `{variable_name}`, but got '{actual_value}'")]
```
ValeryAntopol, line 62: Any time I read it, I want to add "the" before "variable". Not sure if we care about articles here though.

ValeryAntopol (approved):

air/tests/test_module/integration/last_error.rs
mikevoronov: It's used by `f!` macro.

air/src/execution_step/errors.rs
mikevoronov: Usually, error messages have more relaxed rules, but this one is short, so I'll this.

PR #224, opened 2022-02-24
Fix ap trace handler behaviour

### LINKED ISSUE #221: Bug: ap doesn't update new_to_old_pos structure
Trace handler has a special struct called `new_to_old_pos` intended to track correspondes between positions in new data (that are just forming during current execution) and previous and current ones. This struct then is used by `fold` by stream handler to find appropriate to current iterator value states in previous and current data and adjust sliders accordingly. If this states isn't found, then slider will point to some other states in data and merging could be incorrect.

From the other side, there are two sources of values for streams: `call` and `ap`. And the crux of the problem is that  `ap` handler doesn't update `new_to_old_pos` during merging stage.

Consider the following example:

```wasm
 (seq
    (seq
        ;; let's peers be an array of two values [peer_1_id, peer_2_id]
        (call "{set_variable_id}" ("" "") [] peers)
        (fold peers peer
            (par
                (seq
                    (call peer ("" "") [] value)
                    ;; it's crucial to reproduce this bug to add value to stream
                    ;; with help of ap instruction
                    (ap value $stream)
                )
                (next peer)
            )
        )
    )
    ;; join streams on join_1/join_2 peers in such a way that will have different state:
    ;; join_1 $stream: [peer_1_value, peer_2_value]
    ;; join_2 $stream: [peer_2_value, peer_1_value]
    (fold $stream iterator
        ;; here it'll encounter a bug in trace handler, because fold won't shuffle lores in
        ;; the appropriate way
        (par
            (par
                (call "{join_1_id}" ("" "") [iterator])
                (call "{join_2_id}" ("" "") [iterator])
            )
            (next iterator)
        )
    )
)
```

mikevoronov:
Resolves #221.

### DIFF
--- a/crates/air-lib/trace-handler/src/merger/ap_merger.rs
+++ b/crates/air-lib/trace-handler/src/merger/ap_merger.rs
@@ -28,20 +28,29 @@ pub enum MergerApResult {
 
 pub(crate) fn try_merge_next_state_as_ap(data_keeper: &mut DataKeeper) -> MergeResult<MergerApResult> {
     use ExecutedState::Ap;
+    use PreparingScheme::*;
 
     let prev_state = data_keeper.prev_slider_mut().next_state();
     let current_state = data_keeper.current_slider_mut().next_state();
 
-    let ap = match (prev_state, current_state) {
-        (Some(Ap(prev_ap)), _) => prev_ap,
+    match (prev_state, current_state) {
+        (Some(Ap(prev_ap)), Some(_)) => prepare_merge_result(Some(prev_ap), Both, data_keeper),
+        (Some(Ap(prev_ap)), None) => prepare_merge_result(Some(prev_ap), Previous, data_keeper),
         // check that current state is Ap, but it's impossible to use it, because prev_data
         // could not have streams with such generations
-        (None, Some(Ap(_))) => return Ok(MergerApResult::Empty),
-        (None, None) => return Ok(MergerApResult::Empty),
-        (prev_state, current_state) => return Err(MergeError::incompatible_states(prev_state, current_state, "ap")),
-    };
+        (None, Some(Ap(_))) => prepare_merge_result(None, Current, data_keeper),
+        (None, None) => Ok(MergerApResult::Empty),
+        (prev_state, current_state) => Err(MergeError::incompatible_states(prev_state, current_state, "ap")),
+    }
+}
 
-    to_merger_result(ap)
+fn prepare_merge_result(ap_result: Option<ApResult>, scheme: PreparingScheme, data_keeper: &mut DataKeeper) -> MergeResult<MergerApResult> {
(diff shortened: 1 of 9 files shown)


crates/air-lib/trace-handler/src/merger/position_mapping.rs
```diff
@@ -14,33 +14,30 @@
  * limitations under the License.
  */
 
-use super::*;
+use super::DataKeeper;
+use super::DataPositions;
 
-pub(super) enum PrepareScheme {
+pub(super) enum PreparingScheme {
```
ValeryAntopol, line 20: Maybe `Preparation`?

crates/air-lib/trace-handler/src/merger/ap_merger.rs
```diff
@@ -28,20 +28,33 @@ pub enum MergerApResult {
...
+        (Some(Ap(prev_ap)), None) => prepare_merge_result(Some(prev_ap), Previous, data_keeper),
         // check that current state is Ap, but it's impossible to use it, because prev_data
         // could not have streams with such generations
-        (None, Some(Ap(_))) => return Ok(MergerApResult::Empty),
-        (None, None) => return Ok(MergerApResult::Empty),
-        (prev_state, current_state) => return Err(MergeError::incompatible_states(prev_state, current_state, "ap")),
-    };
+        (None, Some(Ap(_))) => prepare_merge_result(None, Current, data_keeper),
+        (None, None) => Ok(MergerApResult::Empty),
+        (prev_state, current_state) => Err(MergeError::incompatible_states(prev_state, current_state, "ap")),
```
ValeryAntopol, line 43: Maybe use named constant instead of "ap"?

ValeryAntopol, line 43: Maybe use named constant instead of "ap"?

crates/air-lib/trace-handler/src/merger/position_mapping.rs
```diff
@@ -14,33 +14,30 @@
...
-        PrepareScheme::Current => None,
+        PreparingScheme::Previous | PreparingScheme::Both => Some(data_keeper.prev_slider().position() - 1),
+        PreparingScheme::Current => None,
     };
 
     let current_pos = match scheme {
-        PrepareScheme::Current | PrepareScheme::Both => Some(data_keeper.current_slider().position() - 1),
-        PrepareScheme::Previous => None,
+        PreparingScheme::Current | PreparingScheme::Both => Some(data_keeper.current_slider().position() - 1),
+        PreparingScheme::Previous => None,
```
ValeryAntopol, line 36: I suggest adding `use PreparingScheme::*` at the beginning, as it will improve readability

ValeryAntopol (approved):

PR #228, opened 2022-03-06
Make fold by streams truly deterministic

mikevoronov:
It's was tech debt after inroducing an async VM that streams were considered in previous sync model. It means that it's supposed that AquaVM starts only when `current_data` come. But after async PR, there is one more reason (namely providing service results) to do that. To make fold by stream deterministic especially in case when `next` in `fold` isn't the last state it's necessary to increase stream generation on each AquaVM launch.

This PR fixes this behaviour and makes fold by stream truly deterministic.

### DIFF
--- a/air/tests/test_module/integration/data_merge.rs
+++ b/air/tests/test_module/integration/data_merge.rs
@@ -18,7 +18,7 @@ use air_test_utils::prelude::*;
 use std::collections::HashMap;
 
 #[test]
-fn data_merge() {
+fn data_merge__() {
     use executed_state::*;
 
     let set_variable_id = "set_variable";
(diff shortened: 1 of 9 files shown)


air/tests/test_module/integration/data_merge.rs
```diff
@@ -18,7 +18,7 @@ use air_test_utils::prelude::*;
 use std::collections::HashMap;
 
 #[test]
-fn data_merge() {
+fn data_merge__() {
```
ValeryAntopol, line 21: I dont really understand why this change is needed

ValeryAntopol (approved):

PR #238, opened 2022-04-04
Abstract over values

mikevoronov:


mikevoronov:
Cancel this PR because it's obsolate and will be split in more granular PRs.

PR #242, opened 2022-04-08
Expose max memory size

mikevoronov:


avm/server/src/runner.rs
```diff
@@ -34,6 +34,15 @@ pub struct AVMRunner {
     wasm_filename: String,
 }
 
+/// Return statistic of AVM server Wasm module heap footprint.
+pub struct AVMMemoryStats {
+    /// Size of currently used linear memory in bytes.
+    /// Please note that linear memory contains not only heap, but globals, shadow stack and so on.
+    pub memory_size: usize,
+    /// Possibly set max memory size for AVM server.
+    pub max_memory_size: Option<usize>,
```
ValeryAntopol, line 43: I dont see places where we can set this value, but comment says that we can

ValeryAntopol (approved):

PR #243, opened 2022-04-14
Improve non iterable scalar scoping

### LINKED ISSUE #241: Bug: scalars in fold aren't cleaned after end of each iteration
AIR supports shadowing of scalars, but they aren't cleaned after end of a fold iteration, and the current situation is possible:

```wat
(fold array-inline-0 node
      (par
            (call node ("kad" "neighborhood") [k-0 [] []] nodes-1) ;; (1)
            (fold nodes-1 n-0 ;; (2)
                  ... ;; (3)
            )
      )
)
```

let's `array-inline-0` contains more then 1 elements and on the second iteration `fold (2)` could iterate over `nodes-1` that was set on the previous one, because `(1)` subtree is called in-par with this `fold`. Along with that execution flow is transmitted to `node` in `(1)` and `fold (2)` could execute for different `nodes-1` array.

As a result, `(3)` become nondeterministic and trace handler could obtain non equal states.

We should discuss whether this bug is AIR one or Aqua, because this bug is only possible when scalar is set in one subtree and used in other, that is prohibited.

mikevoronov:
Resolves #241.


air/tests/test_module/issues/mod.rs
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 Fluence Labs Limited
+ * Copyright 2021 Fluence Labs Limited
```
ValeryAntopol, line 2: maybe 2022?

crates/air-lib/air-parser/src/parser/errors.rs
```diff
@@ -15,34 +15,93 @@
  */
 
 use crate::parser::lexer::LexerError;
+use crate::parser::Span;
+
 use thiserror::Error as ThisError;
 
+// TODO: replace usize pair with Span
```
ValeryAntopol, line 22: looks like this todo is done

air/src/execution_step/execution_context/scalar_variables.rs
```diff
@@ -62,11 +64,32 @@ use std::rc::Rc;
...
 #[derive(Default)]
 pub(crate) struct Scalars<'i> {
-    // this one is optimized for speed (not for memory), because it's unexpected
-    // that a script could have a lot of inner folds.
-    pub values: HashMap<String, Vec<Option<ValueAggregate>>>,
-    pub iterable_values: HashMap<String, FoldState<'i>>,
-    pub fold_block_id: usize,
+    // TODO: use Rc<String> to avoid copying
+    /// Local values could be considered as a sparse matrix, where a raw corresponds to a value
+    /// name and contains all its variants were set with respect to a depth. This structure
```
ValeryAntopol, line 69: Please put "value" and "name" on one line, it will be easier to read. Also "contains all its variants were set with respect to a depth" is a bit cryptic, possibly it is simpler to use "column" to describe this part (you can also keep this sentence, just add something about columns). Also "row", not "raw".

crates/air-lib/air-parser/src/parser/validator.rs
```diff
@@ -50,8 +50,12 @@ pub struct VariableValidator<'i> {
     /// due to the way how lalrpop work.
     unresolved_iterables: MultiMap<&'i str, Span>,
 
+    /// Contains all met iterable in call and next, they will be resolved after the whole parsing
+    /// due to the way how lalrpop work.
+    multiple_next_candidates: MultiMap<&'i str, Span>,
+
     /// Contains all names that should be checked that they are not iterators.
-    check_for_non_iterators: Vec<(&'i str, Span)>,
+    not_iterators_candidates: Vec<(&'i str, Span)>,
```
ValeryAntopol, line 58: maybe `forbidden_iterator_names`?

ValeryAntopol (approved):
LGTM

crates/air-lib/air-parser/src/parser/validator.rs
mikevoronov: Actually, they become forbidden only after postprocessing whether this field contains names that should be processed.

air/src/execution_step/execution_context/scalar_variables.rs
```diff
@@ -119,31 +164,34 @@ impl<'i> Scalars<'i> {
...
-                scalars
-                    .iter()
-                    .take(self.fold_block_id + 1)
-                    .rev()
-                    .find_map(|scalar| scalar.as_ref())
+            .and_then(|values| {
+                let last_cell = values.last();
+                let value_not_invalidated = !self.invalidated_depths.contains(&last_cell.depth);
+
+                if is_global_value(last_cell.depth) || value_not_invalidated {
```
ValeryAntopol, line 177: I don't think that is_global_value is needed, because 0 will never be in invalidated_depths.

ValeryAntopol (approved):

PR #250, opened 2022-04-19
introduce %timestamp%

### LINKED ISSUE #249: Introduce keyword %timestamp%
For proper support of spells in Aqua AIR needs an support of timestamp that could be obtained from a particle. Timestamp is a `u64` value that would be pass in a "standard manner" for the interpreter via `RunParameters`.

mikevoronov:
Resolves #249.


crates/air-lib/interpreter-interface/src/run_parameters.rs
```diff
@@ -27,4 +29,28 @@ pub struct RunParameters {
...
+        }
+    }
+
+    pub fn into_ivalue(self) -> IValue {
+        let run_parameters = vec![
+            IValue::String(self.init_peer_id),
+            IValue::String(self.current_peer_id),
+            IValue::U64(self.timestamp),
+        ];
+        let run_parameters = NEVec::new(run_parameters).unwrap();
```
ValeryAntopol, line 53: do we need a comment about this unwrap?

avm/server/src/runner.rs
```diff
@@ -133,7 +131,7 @@ fn prepare_args(
         IValue::String(air.into()),
         IValue::ByteArray(prev_data.into()),
         IValue::ByteArray(data.into()),
-        IValue::Record(run_parameters),
+        run_parameters,
```
ValeryAntopol, line 134: From the changes i see in this PR, I do not understand why this change is needed or possible

ValeryAntopol (approved):

air/src/execution_step/execution_context/context.rs
```diff
@@ -96,9 +87,29 @@ impl<'i> ExecutionCtx<'i> {
...
 }
 
+use serde::Deserialize;
+use serde::Serialize;
 use std::fmt::Display;
 use std::fmt::Formatter;
 
+/// It reflects RunParameters structure due to limitation of the marine macro to support Rc.
+#[derive(Debug, Default, Clone, Eq, PartialEq, Serialize, Deserialize)]
+pub(crate) struct RcRunParameters {
```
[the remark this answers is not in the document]
mikevoronov: Hmm, don't know tbh, I think that `Rc` prefix makes more sense then `Light` suffix, because one could interpret `Light` differently. Let's stick with `Rc` until we come out with a better idea.

avm/server/src/runner.rs
mikevoronov: Discussed in private.

avm/client/src/importObject.ts
```diff
@@ -0,0 +1,4 @@
+// auto-generated
+
+export const __wbg_callserviceimpl = "";
+export const __wbg_getcurrentpeeridimpl = "";
```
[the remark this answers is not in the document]
mikevoronov: It was changed by accident, thanks.

PR #262, opened 2022-05-13
Use pooled AVMRunner for tests

monoid:
Use pooled `avm::server::AVMRunner` instances of
air_iterpreter_server.wasm to reduce tests' running time.  It avoids
repeated WASM loading and compilation.

On my hardware, `cargo test --release` execution time (precompiled)
decreases from almost 6 minutes to 1.5 minutes.

### DIFF
--- a/avm/server/src/runner.rs
+++ b/avm/server/src/runner.rs
@@ -110,6 +110,11 @@ impl AVMRunner {
             max_memory_size: stats[0].max_memory_size,
         }
     }
+
+    #[inline]
+    pub fn set_peer_id(&mut self, current_peer_id: impl Into<String>) {
+        self.current_peer_id = current_peer_id.into();
+    }
 }
 
 fn prepare_args(
--- a/crates/air-lib/test-utils/Cargo.toml
+++ b/crates/air-lib/test-utils/Cargo.toml
@@ -20,4 +20,6 @@ avm-server = { path = "../../../avm/server" }
 marine-rs-sdk = "0.6.15"
 
 fstrings = "0.2.3"
+object-pool = "0.5.4"
+once_cell = "1.10.0"
 serde_json = "1.0.61"
--- a/crates/air-lib/test-utils/src/test_runner.rs
+++ b/crates/air-lib/test-utils/src/test_runner.rs
@@ -17,16 +17,31 @@
 use super::CallServiceClosure;
 use avm_server::avm_runner::*;
 
+use once_cell::sync::OnceCell;
 use std::collections::HashMap;
 use std::collections::HashSet;
 use std::path::PathBuf;
 
 // 10 Mb
 const AVM_MAX_HEAP_SIZE: u64 = 10 * 1024 * 1024;
+const AIR_WASM_PATH: &str = "../target/wasm32-wasi/debug/air_interpreter_server.wasm";
 
 pub struct TestRunner {
-    pub runner: AVMRunner,
+    pub runner: object_pool::Reusable<'static, AVMRunner>,
     pub call_service: CallServiceClosure,
+}
+
+fn make_pooled_avm_runner() -> AVMRunner {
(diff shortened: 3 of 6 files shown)


avm/server/src/runner.rs
```diff
@@ -110,6 +110,11 @@ impl AVMRunner {
             max_memory_size: stats[0].max_memory_size,
         }
     }
+
+    #[inline]
+    pub fn set_peer_id(&mut self, current_peer_id: impl Into<String>) {
```
monoid, line 115: This method is used in tests only.  Should I make it really public and write tests for it, or make it `hidden`?

crates/air-lib/test-utils/src/test_runner.rs
```diff
@@ -25,10 +26,24 @@ use std::path::PathBuf;
 const AVM_MAX_HEAP_SIZE: u64 = 10 * 1024 * 1024;
 
 pub struct TestRunner {
-    pub runner: AVMRunner,
+    pub runner: object_pool::Reusable<'static, AVMRunner>,
     pub call_service: CallServiceClosure,
 }
 
+fn counctruct_pooled_avm_runner() -> AVMRunner {
```
mikevoronov, line 33: construct

mikevoronov: btw, `construct` looks too long for a name, maybe just create or make?

crates/air-lib/test-utils/src/test_runner.rs
```diff
@@ -97,17 +112,19 @@ pub fn create_avm(
...
     current_peer_id: impl Into<String>,
 ) -> TestRunner {
-    let air_wasm_path = PathBuf::from("../target/wasm32-wasi/debug/air_interpreter_server.wasm");
-    let current_peer_id = current_peer_id.into();
-    let logging_mask = i32::MAX;
+    static POOL_CELL: OnceCell<object_pool::Pool<AVMRunner>> = OnceCell::new();
+
+    let pool = POOL_CELL.get_or_init(|| {
+        object_pool::Pool::new(
+            // We create an empty pool and let it fill on demand.
```
mikevoronov, line 119: According to coding standard used here one-sentence comment starts with non-capital letter and a dot isn't set at the end. Doc comments start with capital letter and ending with a dot.

crates/air-lib/test-utils/src/test_runner.rs
```diff
@@ -25,10 +26,24 @@ use std::path::PathBuf;
 const AVM_MAX_HEAP_SIZE: u64 = 10 * 1024 * 1024;
 
 pub struct TestRunner {
-    pub runner: AVMRunner,
+    pub runner: object_pool::Reusable<'static, AVMRunner>,
     pub call_service: CallServiceClosure,
 }
 
+fn counctruct_pooled_avm_runner() -> AVMRunner {
+    let air_wasm_path = PathBuf::from("../target/wasm32-wasi/debug/air_interpreter_server.wasm");
```
mikevoronov, line 34: Following the boyscout rule, let's introduce a global const for this path.

avm/server/src/runner.rs
mikevoronov: Nope, it doesn't needed, just leave as is.

crates/air-lib/test-utils/Cargo.toml
```diff
@@ -20,4 +20,6 @@ avm-server = { path = "../../../avm/server" }
 marine-rs-sdk = "0.6.15"
 
 fstrings = "0.2.3"
+object-pool = "0.5"
+once_cell = "1.10"
```
mikevoronov, line 24: Suggest to specify patch versions, we've experiences problems with accidentally breaking changes in patch versions of several crates.

mikevoronov (approved):
utACK with one comment

PR #265, opened 2022-05-16
Rename subtree to subgraph

mikevoronov:
Intially, a fold instruction was designed based on idea that its iterable can't expand,
so AIR was considered as a subtree. But after introducing recursive streams in #225, it's
no more a subtree, but a subgraph.


monoid:
Is it a generic graph, or something like a DAG?

mikevoronov:
> Is it a generic graph, or something like a DAG?

It's a generic one because of recursive streams could be considered as a cycle.

ValeryAntopol:
I think that it is also directed, and there exists a node that all other nodes are reachable from it.

mikevoronov:
> I think that it is also directed, and there exists a node that all other nodes are reachable from it.

More precisely, it's a directed rooted graph.

ValeryAntopol (approved):
utACK

Issue #184, opened 2021-11-30
AIR script beautifier


mikevoronov:
AIR code could be represented only in wat-like brackets syntax that hard to read and analize. It's proposed to add a format based on identation and similar to Aqua.

mikevoronov:
I think that we could start with the following syntax transformataion and see how it goes:

```wat
(call 1)
=>
call 1

(ap x y)
=>
ap x y

(seq
  (call 1)
  (call 2)
)
=>
call 1
call 2

(par
  (call 1)
  (call 2)
)
=>
par left:
 call 1
 <~> (maybe it's not the best variant, but let's start with it and see how it looks like)
 call 2

(fold a b
 (seq
   (call 1)
   (next)
 )
)
=>
fold iterable i:
  call 1
  next i

(fold a b
 (par
   (call 1)
   (next)
 )
)
=>
fold iterable i:
  par:
  	call 1
  	<~>
  	next i

(xor
  (call 1)
  (call 2)
)
=> 
try:
 call 1
catch:
 call 2

(new $stream
  (call 1)
)
=>
new $stream:
  call 1

(match lhs rhs
  (call 1)
)
=>
match lhs rhs:
  call 1

(fail x)
=>
fail x
```

folex:
```
(seq
  (call 1)
  (seq
    (call 2)
    (call 3)
  )
)
```
=>
```
call 1
call 2
call 3
```

monoid:
Some implementation details:
1. `(call "a" ("b" "c") [i j k] var)` is formated as `var <- call "a" ("b", "c") [i, j, k])`.
2. The bar in the `par` has same indentation as `par`, for symmetry with `try/catch`:

```
par:
    abc
|
   def
```

PR #266, opened 2022-05-17
`beautifier` and `air-beautify` crates

monoid:
The `air-beautify` util takes an AIR script input with Lisp-like syntax
and transforms it into experimental indentation-based syntax.

Closing #184.


crates/beautifier/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+license = "Apache-2.0"
+repository = "https://github.com/fluencelabs/aquavm/tree/master/crates/beautifier"
+publish = false
+keywords = ["fluence", "air", "programming-language"]
+
+[lib]
+name = "air_beautifier"
+path = "src/lib.rs"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
mikevoronov, line 16: ditto

crates/beautifier/src/lib.rs
```diff
@@ -0,0 +1,292 @@
...
+pub enum BeautifyError {
+    #[error("{0}")]
+    Parse(String),
+    #[error(transparent)]
+    Io(#[from] IoError),
+}
+
+/**
+ * AIR beautifier.
+*/
```
mikevoronov, line 109: In this repo we use the following format for doc comments:
```
/// This is a doc comment with dot at the end.
```

crates/beautifier/src/lib.rs
```diff
@@ -0,0 +1,292 @@
...
+    nonstandard_style,
+    unused_imports,
+    unused_mut,
+    unused_variables,
+    unused_unsafe,
+    unreachable_patterns
+)]
+
+use std::fmt::Display;
+use std::io::{Error as IoError, Result as IoResult, Write};
```
mikevoronov, line 28: I think that `Write` is too broader name to be exported to the global scope, suggest to use it with the nearest mod prefix: `io::Write`.

crates/beautifier/src/lib.rs
```diff
@@ -0,0 +1,292 @@
...
+
+fn fmt_indent(output: &mut impl Write, indent: usize) -> IoResult<()> {
+    // basic implementation that can be speeded up
+    for _ in 0..indent {
+        write!(output, " ")?;
+    }
+    Ok(())
+}
+
+struct BArgs<'a, 'b>(&'a [Value<'b>]);
```
mikevoronov, line 69: Let's use more meaningful names for lifetimes

crates/beautifier/src/lib.rs
```diff
@@ -0,0 +1,292 @@
...
+            } else {
+                first = false;
+            }
+            write!(f, "{}", item)?;
+        }
+        Ok(())
+    }
+}
+
+struct BTriplet<'a, 'b>(&'a Triplet<'b>);
```
mikevoronov, line 86: ditto

crates/beautifier/src/lib.rs
```diff
@@ -0,0 +1,292 @@
...
+    unused_mut,
+    unused_variables,
+    unused_unsafe,
+    unreachable_patterns
+)]
+
+use std::fmt::Display;
+use std::io::{Error as IoError, Result as IoResult, Write};
+
+use air_parser::ast::{
```
mikevoronov, line 30: In this repo we follow this import scheme:
- imports from local project (`crate::`/`super::`) 
- newline
- other imports from non-std
- newline
- imports from std

If there are not so many imports newlines could be omitted.

mikevoronov: Do we need to import all instructions to the global scope? Maybe just import `air_parser::ast` and then use `ast` as a prefix?

crates/beautifier/src/lib.rs
```diff
@@ -0,0 +1,292 @@
...
+        {
+            let out = &mut $beautifier.output;
+            $crate::fmt_indent(out, $indent)?;
+            writeln!(out, $fmt2 $(, $arg2)*)?;
+        }
+        crate::Beautifier::beautify_walker(&mut *$beautifier, $nest2, $indent + step)
+    });
+}
+
+fn fmt_indent(output: &mut impl Write, indent: usize) -> IoResult<()> {
```
mikevoronov, line 61: You could use the following approach for indentation:
```rust
write!(w, "{:indent$}Indented text!", "", indent=indent)
```

[playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&gist=ab951cb07db3b1415f048341eeda42f7)

crates/beautifier/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "beautifier"
```
mikevoronov, line 2: Why package name is `beautifier`, but lib name is `air_beautifier`?

air-beautify/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
```
mikevoronov, line 1: Suggest to put it into `tools/cli` dir as we discussed.

air-beautify/src/main.rs
```diff
@@ -0,0 +1,82 @@
...
+    let mut air_script = String::new();
+
+    match &args.input {
+        Some(in_path) => {
+            let mut file = File::open(in_path)?;
+            file.read_to_string(&mut air_script)?;
+        }
+        None => {
+            let mut stdin = io::stdin().lock();
+            stdin.read_to_string(&mut air_script)?;
```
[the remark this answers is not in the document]
monoid: It reads until EOF.

air-beautify/src/main.rs
```diff
@@ -0,0 +1,82 @@
...
+    unused_unsafe,
+    unreachable_patterns
+)]
+
+use air_beautifier::Beautifier;
+use anyhow::{Context, Result};
+use clap::Parser;
+use std::{
+    fs::File,
+    io::{self, Read},
```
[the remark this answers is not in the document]
monoid: Do you suggest importing it within a function?
I always prefer global imports.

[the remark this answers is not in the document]
mikevoronov: Yep, I suggest to import trait that is used only in one function locally.

crates/beautifier/src/lib.rs
monoid: What are they?  The `Value` uses `'i` in its declaration.  Can't say I do understand its meaning.

crates/beautifier/src/lib.rs
```diff
@@ -0,0 +1,292 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let mut first = true;
+        for item in self.0 {
+            if !first {
+                write!(f, ", ")?;
+            } else {
+                first = false;
+            }
+            write!(f, "{}", item)?;
+        }
```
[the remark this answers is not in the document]
monoid: Found `.format` in the itertools.

crates/beautifier/src/lib.rs
mikevoronov: `'i` is dervied from `instruction`, because they are from parsed instructions. `'a` could be named `'ctx`

crates/beautifier/src/lib.rs
```diff
@@ -0,0 +1,264 @@
...
+    dead_code,
+    nonstandard_style,
+    unused_imports,
+    unused_mut,
+    unused_variables,
+    unused_unsafe,
+    unreachable_patterns
+)]
+
+use air_parser::ast;
```
mikevoronov, line 27: I suggest to move all logic from `lib.rs` to a separate `beautifier.rs` file and add here only necessary reexports + two more functions
`beautify(air_script: &str, &mut impl io::Write) -> IoResult<()>`,
`beautify_to_string` - it'll acts exactly as one in tests, and in the future we could export it in the Wasm side

crates/beautifier/src/lib.rs
```diff
@@ -0,0 +1,264 @@
...
+use itertools::Itertools as _;
```
mikevoronov, line 28: let's export it locally

crates/beautifier/src/lib.rs
mikevoronov: What about `io::Write`? It is still exported globally.

crates/beautifier/src/lib.rs
```diff
@@ -0,0 +1,264 @@
...
+        $crate::Beautifier::beautify_walker($beautifier, $nest1, $indent + $beautifier.indent_size)
+    });
+    ($beautifier:expr, $indent:expr, $fmt1:literal $(, $arg1:expr)*; $nest1:expr; $fmt2:literal $(, $arg2:expr)*; $nest2:expr) => ({
+        let step = $beautifier.indent_size;
+        {
+            let out = &mut $beautifier.output;
+            $crate::fmt_indent(out, $indent)?;
+            writeln!(out, $fmt1 $(, $arg1)*)?;
+        }
+        crate::Beautifier::beautify_walker(&mut *$beautifier, $nest1, $indent + step)?;
```
mikevoronov, line 49: Can we introduce a new pattern in this macro to avoid code duplication here?

mikevoronov: Or we could just call the first pattern here

monoid: I thought about rewriting it into single-branch macro with repetition pattern.

crates/beautifier/src/lib.rs
monoid: Will do as the last refactoring.

crates/beautifier/src/lib.rs
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    fn beautify_next(&mut self, next: &ast::Next, indent: usize) -> IoResult<()> {
+        fmt_indent(&mut self.output, indent)?;
+        writeln!(&mut self.output, "next {}", next.iterator.name)
+    }
+
+    fn beautify_null(&mut self, null: &ast::Null, indent: usize) -> IoResult<()> {
+        fmt_indent(&mut self.output, indent)?;
+        // emits correct text
```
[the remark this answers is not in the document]
monoid: Removed it at all.

crates/beautifier/src/lib.rs
```diff
@@ -0,0 +1,264 @@
...
+        )
+    }
+
+    fn beautify_ap(&mut self, ap: &ast::Ap, indent: usize) -> IoResult<()> {
+        fmt_indent(&mut self.output, indent)?;
+        writeln!(&mut self.output, "{}", ap)
+    }
+
+    fn beautify_seq(&mut self, seq: &ast::Seq, indent: usize) -> IoResult<()> {
+        // please note that seq uses same indendation intentionally
```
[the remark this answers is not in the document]
monoid: Removed the text.

crates/beautifier/src/beautifier.rs
```diff
@@ -0,0 +1,228 @@
...
+    /// Beautifier for the output with default indent size.
+    pub fn new(output: W) -> Self {
+        Self {
+            output,
+            indent_size: DEFAULT_INDENT_SIZE,
+        }
+    }
+
+    /// Beautifier for the output with custom indent size.
+    pub fn new_with_indent(output: W, indent_step: usize) -> Self {
```
mikevoronov, line 111: We have several similar terms in this file:
`indent_step`,
`step`,
`indent`,
`indent_size`,

Can we reduce their count somehow?

mikevoronov (approved):
utACK

PR #269, opened 2022-05-19
Refactor `call` triplets

### LINKED ISSUE #267: Refactor `call` triplets
`Call` instruction supports three valid variants of a triplet `(peer_id service_id) function_name`, `peer_id (service_id function_name)`, `(peer_id service_id) (service_id function_name)`. It was made to simplify developing on AIR scripts, but after introducing Aqua it's no more needed, now Aqua generates only one variant `peer_id (service_id function_name)` and only this variant is used across AquaVM tests.

monoid:
Older grammar allowed some variabilty in triple format.  Now, the only format allowed is:

  `peer_id (service_id function_name)`

As less errors can happen during parsing of the triplet, the code was
further simplified.

Closes #267.


mikevoronov (approved):
utACK

PR #270, opened 2022-05-30
Set an empty subtrace on absent fold lore

### LINKED ISSUE #222: Bug: fold FSM handles empty states in incorrect way
Trace handler has sliders that is used to point on appropriate states in current and previous datas. These sliders are adjusted correspondingly to the fold lores (it's internal name of data states descriptors), and in its turn lores are chosen by value id that `fold` currently iterates on. 

The problem is `fold` handler doesn't do anything in case of absence of value in lores. This could cause undeterministic execution and errors in state merging process.

This issue is close to #221, they were split in order to better fixing.

monoid:
Closes #222

### DIFF
--- a/air/tests/test_module/issues/issue_222.rs
+++ b/air/tests/test_module/issues/issue_222.rs
@@ -20,65 +20,60 @@ use pretty_assertions::assert_eq;
 
 #[test]
 // test for github.com/fluencelabs/aquavm/issues/222
-fn issue_222_possibly() {
-    let air_script = r#"
-(new $stream
-  (par
-   (par
-      (call "other_1" ("" "") [] $stream)
-      (call "other_2" ("" "") [] $stream))
-   (fold $stream j
-       (seq (call "other_id" ("" "") [j])
-            (next j)))))
-"#;
+fn issue_222_weak() {
+    let other_id = "other_id";
+    let other_1_id = "other_1";
+    let other_2_id = "other_2";
 
-    let mut other_id = create_avm(echo_call_service(), "other_id");
+    let air_script = f!(r#"
+        (new $stream
+          (par
+           (par
+              (call "{other_1_id}" ("" "") [] $stream)
+              (call "{other_2_id}" ("" "") [] $stream))
+           (fold $stream j
+               (seq (call "{other_id}" ("" "") [j])
+                    (next j)))))
+"#);
 
-    // The bug is triggered when (call "other_2" ...) result arrives to "other_id"
+    let mut other_id_vm = create_avm(echo_call_service(), "other_id");
+    let mut other_1_vm = create_avm(set_variable_call_service(json!([1])), "other_1");
+    let mut other_2_vm = create_avm(set_variable_call_service(json!([2])), "other_2");
+
+    let result = checked_call_vm!(other_id_vm, <_>::default(), &air_script, "", "");
+    let other_1_result = checked_call_vm!(other_1_vm, <_>::default(), &air_script, "", result.data.clone());
+    let other_2_result = checked_call_vm!(other_2_vm, <_>::default(), &air_script, "", result.data.clone());
+
+    // the bug is triggered when (call "other_2" ...) result arrives to "other_id"
     // before the "other_1" result.
-    let cur_data = br#"
-{"trace":[
-  {"par":[3,2]},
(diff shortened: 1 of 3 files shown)


air/tests/test_module/issues/issue_222.rs
```diff
@@ -0,0 +1,85 @@
...
+  (par
+   (par
+      (call "other_1" ("" "") [] $stream)
+      (call "other_2" ("" "") [] $stream))
+   (fold $stream j
+       (seq (call "other_id" ("" "") [j])
+            (next j)))))
+"#;
+
+    let mut other_id = create_avm(echo_call_service(), "other_id");
```
mikevoronov, line 35: I usually add a suffix `vm` to all such variables.

crates/air-lib/trace-handler/src/state_automata/fold_fsm/lore_applier.rs
```diff
@@ -44,16 +44,20 @@ fn apply_fold_lore(
...
     let slider = match ctx_type {
         Previous => data_keeper.prev_slider_mut(),
         Current => data_keeper.current_slider_mut(),
     };
 
+    let fold_lore = match fold_lore {
+        Some(fold_lore) => fold_lore,
+        None => {
+            // create an empty subtrace
+            slider.set_position_and_len(slider.subtrace_len(), 0)?;
```
mikevoronov, line 56: Using `slider.subtrace_len()` for a position is kinda confusing. From the other hand `slider` has a method `set_subtrace_len`, I suggest to use it instead.

crates/air-lib/trace-handler/src/state_automata/fold_fsm/lore_applier.rs
```diff
@@ -44,16 +44,20 @@ fn apply_fold_lore(
...
-        Some(fold_lore) => fold_lore,
-        None => return Ok(()),
-    };
-
     let slider = match ctx_type {
         Previous => data_keeper.prev_slider_mut(),
         Current => data_keeper.current_slider_mut(),
     };
 
+    let fold_lore = match fold_lore {
```
mikevoronov, line 52: I think that `if ...is_none()` would look better.

monoid: We need value in the `Some` anyway.  I've refactored into a `match` statement instead of an expression.

air/tests/test_module/issues/issue_222.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    // The bug is triggered when (call "other_2" ...) result arrives to "other_id"
```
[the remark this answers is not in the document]
monoid: Yep, it was my idea too.

crates/air-lib/trace-handler/src/state_automata/fold_fsm/lore_applier.rs
```diff
@@ -44,29 +44,31 @@ fn apply_fold_lore(
...
-                fold_lore.before_subtrace.begin_pos as _,
-                fold_lore.before_subtrace.subtrace_len as _,
-            )?;
-        }
-        After => {
-            slider.set_position_and_len(
-                fold_lore.after_subtrace.begin_pos as _,
-                fold_lore.after_subtrace.subtrace_len as _,
-            )?;
+    match fold_lore {
```
mikevoronov, line 52: I'd say that `if` was better because it decouples one indentation level and the code becomes more understandable.

air/tests/test_module/issues/issue_222.rs
```diff
@@ -0,0 +1,80 @@
...
+    let air_script = f!(r#"
+        (new $stream
+          (par
+           (par
+              (call "{other_1_id}" ("" "") [] $stream)
+              (call "{other_2_id}" ("" "") [] $stream))
+           (fold $stream j
+               (seq (call "{other_id}" ("" "") [j])
+                    (next j)))))
+"#);
```
mikevoronov, line 37: please indent it also

mikevoronov (commented):
utACK

mikevoronov (approved):

monoid:
I've concluded that the test in this PR is not weak, it perfectly demonstrates the problem.  I've renamed it.

PR #274, opened 2022-06-07
Run tests with sanitizers

### LINKED ISSUE #247: Use sanitizers in CI
Although `unsafe` in AquaVM is used only in certain places (and mainly for creating lambdas), it's still actual to integrate sanitizers in tests at least for detecting possible memory leaks. The good starting point could be this [article](https://doc.rust-lang.org/beta/unstable-book/compiler-flags/sanitizer.html).

monoid:
As sanitizers are not available for WASM yet, AquaVM is run as native code with `--feature test_with_native_code` for these tests.

Closes #247.

### DIFF
--- a/crates/air-lib/test-utils/src/native_test_runner.rs
+++ b/crates/air-lib/test-utils/src/native_test_runner.rs
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+use crate::test_runner::AirRunner;
 use air_interpreter_interface::RunParameters;
 use avm_server::avm_runner::*;
 use avm_server::into_raw_result;
@@ -22,14 +23,14 @@ pub struct NativeAirRunner {
     current_peer_id: String,
 }
 
-impl NativeAirRunner {
-    pub fn new(current_peer_id: impl Into<String>) -> Self {
+impl AirRunner for NativeAirRunner {
+    fn new(current_peer_id: impl Into<String>) -> Self {
         Self {
             current_peer_id: current_peer_id.into(),
         }
     }
 
-    pub fn call(
+    fn call(
         &mut self,
         air: impl Into<String>,
         prev_data: impl Into<Vec<u8>>,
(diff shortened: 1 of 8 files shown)


crates/air-lib/test-utils/src/native_test_runner.rs
```diff
@@ -0,0 +1,87 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+use super::CallServiceClosure;
+use air_interpreter_interface::RunParameters;
+use avm_server::avm_runner::*;
+
+// Borrowed from private module in the avm/server/src/interface/call_service_result.rs
```
mikevoronov, line 21: Let's make it public?

mikevoronov, line 21: Let's make it public?

crates/air-lib/test-utils/src/native_test_runner.rs
```diff
@@ -0,0 +1,87 @@
...
+        let outcome = RawAVMOutcome::from_interpreter_outcome(outcome)?;
+
+        Ok(outcome)
+    }
+}
+
+pub struct TestRunner {
+    pub runner: NativeAirRunner,
+    pub call_service: CallServiceClosure,
+}
```
mikevoronov, line 77: Can we abstract runner and have only one definition of `TestRunner`?

mikevoronov, line 77: Can we abstract runner and have only one definition of `TestRunner`?

crates/air-lib/test-utils/src/native_test_runner.rs
monoid: Well, I wasn't aware it is re-exported, so, making it public is enough.

crates/air-lib/test-utils/src/native_test_runner.rs
monoid: Indeed.

PR #277, opened 2022-06-22
Add API to save anomaly data

mikevoronov:
This PR adds two new methods in `DataStore` to determine anomaly and collect necessary data. Additionally
  - a generic parameter of `DataStore` turned to be associative
  - `AVMOutcome` contains additionally `memory_delta` and `execution_time`

### DIFF
--- a/avm/server/src/avm.rs
+++ b/avm/server/src/avm.rs
@@ -108,7 +108,7 @@ impl<E> AVM<E> {
 
         // persist resulted data
         self.data_store.store_data(&outcome.data, particle_id)?;
-        let outcome = AVMOutcome::from_raw_outcome(outcome)?;
+        let outcome = AVMOutcome::from_raw_outcome(outcome, memory_delta, execution_time)?;
 
         Ok(outcome)
     }
--- a/avm/server/src/avm.rs
+++ b/avm/server/src/avm.rs
@@ -85,24 +85,35 @@ impl<E> AVM<E> {
         let current_data = data.into();
 
         let execution_start_time = Instant::now();
+        let memory_size_before = self.runner.memory_stats().memory_size;
         let outcome = self
             .runner
             .call(
                 air,
                 prev_data,
                 current_data.clone(),
-                particle_parameters.init_peer_id.into_owned(),
+                particle_parameters.init_peer_id.clone().into_owned(),
                 particle_parameters.timestamp,
                 particle_parameters.ttl,
                 call_results,
             )
             .map_err(AVMError::RunnerError)?;
 
-        let execution_duration = execution_start_time.elapsed();
-        if self.data_store.should_collect_data(execution_duration) {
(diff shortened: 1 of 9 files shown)


avm/server/src/avm.rs
```diff
@@ -81,20 +83,29 @@ impl<E> AVM<E> {
...
+                current_data.clone(),
+                particle_parameters.init_peer_id.clone().into_owned(),
                 particle_parameters.timestamp,
                 particle_parameters.ttl,
                 call_results,
             )
             .map_err(AVMError::RunnerError)?;
 
+        let execution_time = execution_start_time.elapsed();
+        let memory_delta = self.runner.memory_stats().memory_size - memory_size_before;
```
ValeryAntopol, line 104:
```suggestion
        let memory_delta = self.memory_stats().memory_size - memory_size_before;
```

ValeryAntopol (commented):
utACK

crates/data-store/src/lib.rs
```diff
@@ -14,13 +14,35 @@
...
+
+    /// Cleanup data that become obsolete.
+    fn cleanup_data(&mut self, key: &str) -> Result<(), Self::Error>;
+
+    /// Returns true if an anomaly happened and it's necessary to save execution data
+    /// for debugging purposes.
+    ///  execution_time - is a time taken by the interpreter to execute provided script
+    ///  memory_delta - is a count of bytes on which an interpreter heap has been extended
+    ///                 during execution of a particle
+    fn detect_anomaly(&self, execution_time: Duration, memory_delta: usize) -> bool;
```
monoid, line 39: It might be bit disputable to put anomaly detection logic into a DataStore.

avm/server/src/avm.rs
```diff
@@ -81,23 +83,32 @@ impl<E> AVM<E> {
...
+        let execution_start_time = Instant::now();
+        let memory_size_before = self.memory_stats().memory_size;
         let outcome = self
             .runner
             .call(
                 air,
                 prev_data,
-                data,
-                particle_parameters.init_peer_id.into_owned(),
+                current_data.clone(),
```
monoid, line 95: I have lot of ideas to discuss looking at this line.

crates/data-store/src/lib.rs
mikevoronov: I had the same thoughts and we dicsussed it with @folex and it looks like node code would become too complicated after introducing a new boxed trait to AVM. So we decided to do that in this trait.

folex (approved):
Reading `prev_data` from disk in `save_anomaly_data` looks a little bit error-prone: gotta be careful to not overwrite it with `new_data` before `save_anomaly_data` is called. But I understand the motivation.

So utACK :)

folex (approved):
Tested ACK

PR #279, opened 2022-06-24
Performance metrics for AquaVM  based on the `tracing` crate

monoid:
Current version outputs JSON traces with execution times of important parts of the AquaVM.


air-interpreter/src/logger.rs
```diff
@@ -15,7 +15,6 @@
  */
 
 use air_log_targets::TARGET_MAP;
-
```
ValeryAntopol, line 18: Its nice to have internal/external crates imports separated, so lets keep the blank line

air-interpreter/src/wasm_bindgen.rs
```diff
@@ -41,8 +41,10 @@ pub const DEFAULT_LOG_LEVEL: LevelFilter = LevelFilter::Trace;
 pub fn main() {
     // it's necessary to initialize it with the minimal allowed log level,
     // because otherwise it's impossible to set less level than used during initialization.
+
```
ValeryAntopol, line 44: Looks like we forgot to remove this compilation target. Maybe we can just remove it in this PR, @mikevoronov what do you think?

ValeryAntopol (commented):
utACK
But i didn't check that set of skipped/traced functions is correct,  @mikevoronov will do a better job looking at it.

air-interpreter/src/wasm_bindgen.rs
monoid: I think removing this file is too important to be merged with other PR.

air/src/execution_step/air/call/resolved_call.rs
```diff
@@ -129,7 +130,11 @@ impl<'i> ResolvedCall<'i> {
             tetraplets,
         } = self.resolve_args(exec_ctx)?;
 
-        let serialized_tetraplets = serde_json::to_string(&tetraplets).expect("default serializer shouldn't fail");
+        let serialized_tetraplets = {
+            let span = tracing::span!(tracing::Level::INFO, "serde_json::to_string(tetraplets)");
```
monoid, line 134: Such construction now exist in few places, and it looks like a great application for a macros.  I haven't found proper place for it.

tools/cli/air-trace/src/utils.rs
```diff
@@ -0,0 +1,38 @@
...
+use std::time::Duration;
+
+pub(crate) fn parse_tracing_duration(input: &str) -> Result<Duration, anyhow::Error> {
+    for (suffix, scale) in [("ns", 1e-9), ("µs", 1e-6), ("ms", 1e-3), ("s", 1e0)] {
+        if let Some(num_str) = input.strip_suffix(suffix) {
+            if let Ok(num) = num_str.parse::<f64>() {
+                return Ok(Duration::from_secs_f64(num * scale));
+            } else {
+                break;
+            }
```
mikevoronov, line 25: Let's use `match`

tools/cli/air-trace/src/utils.rs
```diff
@@ -0,0 +1,38 @@
...
+ */
+use std::time::Duration;
+
+pub(crate) fn parse_tracing_duration(input: &str) -> Result<Duration, anyhow::Error> {
+    for (suffix, scale) in [("ns", 1e-9), ("µs", 1e-6), ("ms", 1e-3), ("s", 1e0)] {
+        if let Some(num_str) = input.strip_suffix(suffix) {
+            if let Ok(num) = num_str.parse::<f64>() {
+                return Ok(Duration::from_secs_f64(num * scale));
+            } else {
+                break;
```
mikevoronov, line 24: Can we use `return Err` here?

tools/cli/air-trace/src/utils.rs
```diff
@@ -0,0 +1,38 @@
...
+    }
+    return Err(anyhow::anyhow!("malformed duration {:?}", input));
+}
+
+pub(crate) fn unix_timestamp_now() -> u64 {
+    use std::time::SystemTime;
+
+    SystemTime::now()
+        .duration_since(SystemTime::UNIX_EPOCH)
+        .expect("time befor Unix epoch")
```
mikevoronov, line 36: befor**e**

tools/cli/air-trace/src/utils.rs
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+use std::time::Duration;
+
+pub(crate) fn parse_tracing_duration(input: &str) -> Result<Duration, anyhow::Error> {
```
mikevoronov, line 18: Are there any dedicated crate for that? Looks like smth like https://docs.rs/parse_duration/latest/parse_duration/ could suit (but I haven't look much to its API).

air/src/execution_step/air/call/resolved_call.rs
mikevoronov: Yep, I'd agree that it's a good application for macros.

air/src/farewell_step/outcome.rs
```diff
@@ -100,7 +113,8 @@ fn populate_outcome_from_contexts(
 }
 
 /// Deduplicate values in a supplied vector.
-fn dedup<T: Eq + Hash>(mut vec: Vec<T>) -> Vec<T> {
+#[tracing::instrument]
```
mikevoronov, line 116: I don't need we actually need to instrument it, other functions will dominate over this deduplication.

air/src/runner.rs
```diff
@@ -64,7 +64,13 @@ fn execute_air_impl(
...
     // match here is used instead of map_err, because the compiler can't determine that
     // they are exclusive and would treat exec_ctx and trace_handler as moved
-    match air.execute(&mut exec_ctx, &mut trace_handler) {
+    match {
+        // create a span for tracing the whole air.execute and its handling
+        let execute_span = tracing::span!(tracing::Level::INFO, "execute");
+        let _enter = execute_span.enter();
+
+        air.execute(&mut exec_ctx, &mut trace_handler)
+    } {
```
mikevoronov, line 73: Let's add a new variable and then match on it.

tools/cli/air-trace/src/run.rs
```diff
@@ -0,0 +1,163 @@
...
+    #[clap(name = "--plain")]
+    PlainData(self::data::plain::PlainDataArgs),
+}
+
+pub(crate) fn run(args: Args) -> anyhow::Result<()> {
+    let tracing_json = (!args.json) as u8;
+    init_tracing(args.tracing_params.clone(), tracing_json);
+
+    let current_peer_id = args.current_peer_id.as_deref().unwrap_or("some_id");
+    let mut runner = if cfg!(not(feature = "wasm")) || args.native {
```
mikevoronov, line 73: Suggest to introduce a dedicated function for getting runner

tools/cli/air-trace/src/run.rs
```diff
@@ -0,0 +1,163 @@
...
+
+use air_test_utils::CallResults;
+
+use anyhow::Context as _;
+use clap::{Parser, Subcommand};
+use std::path::{Path, PathBuf};
+
+#[derive(Parser, Debug)]
+#[clap(about = "Run AIR script with AquaVM")]
+pub(crate) struct Args {
```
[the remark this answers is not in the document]
monoid: Newlines separate logical groups of arguments.  Perhaps, one should refactor out them into flattened structs.

air/src/runner.rs
monoid: It will be solved by the macro.

tools/cli/air-trace/src/utils.rs
monoid: I'v tried several ones, but one does not support floats, another does not support suffixes used by the output format, etc.

mikevoronov (approved):
tested ACK

PR #280, opened 2022-06-30
Add AIR script to the AquaVM's anomaly data

monoid:
Add it to the particle_data, where it belongs logically.

### DIFF
--- a/avm/server/src/avm.rs
+++ b/avm/server/src/avm.rs
@@ -26,6 +26,7 @@ use crate::interface::ParticleParameters;
 use crate::AVMResult;
 
 use avm_data_store::AnomalyData;
+use serde::Serialize;
 use std::ops::Deref;
 use std::ops::DerefMut;
 use std::time::Duration;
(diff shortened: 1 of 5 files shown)


crates/data-store/src/lib.rs
```diff
@@ -49,6 +49,7 @@ pub trait DataStore {
```
mikevoronov, line 49: Bump the `DataStore` minor version also.

avm/server/src/avm.rs
```diff
@@ -144,14 +147,15 @@ impl<E> AVM<E> {
             .data_store
             .read_data(particle_parameters.particle_id.as_str())?;
         let ser_particle =
-            serde_json::to_vec(particle_parameters).map_err(AVMError::AnomalyDataSeError)?;
+            serde_json::to_vec(&particle_parameters).map_err(AVMError::AnomalyDataSeError)?;
```
mikevoronov, line 150: `&` doesn't needed here

mikevoronov (approved):
utACK

PR #281, opened 2022-06-30
Migrate ait workflow to github-actions

nahsi:


.github/workflows/air.yml
```diff
@@ -0,0 +1,58 @@
+name: "air"
```
mikevoronov, line 1: Let's call workflow and action name as `aquavm`.

mikevoronov (approved):
utACK

PR #282, opened 2022-07-06
Add a workaround for wasm memory leak

ValeryAntopol:


mikevoronov (approved):
utACK.

don't forget about formatting.

PR #283, opened 2022-07-08
AquaVM benchmarks

monoid:
Fix old ones and add new ones, including important data cases (large values or long traces).


mikevoronov (approved):
tested ACK

PR #285, opened 2022-07-12
Introduce `tracing_benches.sh` script

monoid:
It accepts two arguments: `big`/`long` test type and `merge`/`nomerge`
test subtype.  Any of arguments can be replaced with `--all` value.

### DIFF
--- a/air/tracing_benches.sh
+++ b/air/tracing_benches.sh
@@ -68,9 +68,9 @@ function main {
     esac
     shift 2
 
-    : "${AIR_WASM_RUNTIME_PATH:=../target/wasm32-wasi/release/air_interpreter_server.wasm}"
-    if [ ! -e "${AIR_WASM_RUNTIME_PATH}" ]; then
-        error "No AIR interpreter at ${AIR_WASM_RUNTIME_PATH}."
+    : "${AIR_INTERPRETER_WASM_PATH:=../target/wasm32-wasi/release/air_interpreter_server.wasm}"
+    if [ ! -e "${AIR_INTERPRETER_WASM_PATH}" ]; then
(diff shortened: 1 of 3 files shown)


air/tracing_benches.sh
```diff
@@ -0,0 +1,123 @@
...
+        merge|nomerge)
+            subtests=( "$2" )
+            ;;
+        --all)
+            subtests=( merge nomerge )
+            ;;
+        *)
+            error "Unknown subtest type: $2"
+            ;;
+    esac
```
ValeryAntopol, line 68: Incorrect parentheses sequence drives me mad. Maybe it is correct syntax that i don't know, but it is quite hard to understand. Not a request for change, just a notice.

air/tracing_benches.sh
```diff
@@ -0,0 +1,123 @@
...
+    shift 2
+
+    : "${AIR_WASM_RUNTIME_PATH:=../target/wasm32-wasi/release/air_interpreter_server.wasm}"
```
ValeryAntopol, line 71: Maybe `AIR_INTERPRETER_WASM`? Don't like the `runtime` word here. Also the binary name is `interpreter`, not `runtime`

ValeryAntopol (approved):
utACK

air/tracing_benches.sh
monoid: This is variable used by the air-trace, and is to be renamed with the related option.

ValeryAntopol: Yep. Let's change it now in both script and air-trace, or late in a separate PR

air/tracing_benches.sh
monoid: Unfortunately, this is bash's official `case` syntax that has no alternative (except using `if`). Just an informational message :)

PR #286, opened 2022-07-15
Fix rustdoc generation

monoid:
It fails because of rust-lang/rust#99261 at nightly-2022-07-14.
Enforce nightly-2022-07-13.


mikevoronov (approved):
utACK

monoid:
The it has been fixed upstream, the PR is unnecessary.

PR #289, opened 2022-08-04
Adjust air-trace options

monoid:
1. Introduce --runner-tracing-params for enable and disable tracing of runner itself.
2. Adjust some functions' tracing levels to make default traces less noisy.


mikevoronov (approved):
utACK

PR #290, opened 2022-08-08
Remove `jsonpath` filter option

mikevoronov:
`jsonpath` has been used in early versions of AquaVM as lambda implementation, then it was replaced by our own simple lambda implementation, because of lack of standard and canonical realization. This filter option was forgotten to be deleted.


ValeryAntopol (approved):
utACK

PR #291, opened 2022-08-08
Fix clippy warnings

mikevoronov:


tools/cli/air-trace/src/stats.rs
```diff
@@ -48,6 +48,7 @@ pub(crate) fn stats(mut args: Args) -> anyhow::Result<()> {
 
     let mut stats = self::report::StatsReport::new();
 
+    #[allow(clippy::significant_drop_in_scrutinee)]
```
monoid, line 51: One can avoid this warning by passing `&mut stdin` instead (making `mut stdin` above).
I don't see it in my installation, BTW.

crates/air-lib/air-parser/src/ast/instruction_arguments.rs
```diff
@@ -25,7 +25,7 @@ use air_lambda_ast::LambdaAST;
 use serde::Deserialize;
 use serde::Serialize;
 
-#[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
```
monoid, line 28: Is adding `Eq` solves some warning, or "just in case"?

monoid:
utACK

monoid (approved):

crates/air-lib/air-parser/src/ast/instruction_arguments.rs
mikevoronov: Yep, it solves `clippy` warning.

tools/cli/air-trace/src/stats.rs
mikevoronov: Looks like the reason here is `anyhow::Result`.

monoid: The lint seems to be broken:
https://github.com/rust-lang/rust-clippy/issues/8987

PR #294, opened 2022-08-17
Refactor `avm-server` interface mod to new crate

monoid:
It allows to build a fully native `air-trace` build, including WASM build.


avm/server/src/errors.rs
```diff
@@ -73,38 +73,16 @@ pub enum RunnerError {
...
-
-    /// Errors encountered while trying to deserialize tetraplets from call parameters returned
-    /// by the interpreter. In the corresponding struct such tetraplets are
-    /// Vec<Vec<SecurityTetraplet>> serialized to a string.
-    #[error("error occurred while deserialization of tetraplets from call params `{call_params:?}`: {de_error}")]
-    CallParamsTetrapletsDeFailed {
-        call_params: air_interpreter_interface::CallRequestParams,
-        de_error: SerdeError,
-    },
+// TODO same variant? it will make the RunnerError type little larger; but it is yet to be measured
```
ValeryAntopol, line 76: That does not necessary will make this type larger. Rust aggressively optimizes enums for space, as described here https://www.sobyte.net/post/2022-04/rust-enum-layout/. So maybe its ok to just embed  AVMInterfaceError in RunnerError

ValeryAntopol (approved):
utACK

avm/interface/src/lib.rs
```diff
@@ -0,0 +1,77 @@
+/*
+ * Copyright 2021 Fluence Labs Limited
```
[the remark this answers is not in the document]
monoid: That's just moved `mod.rs` file...

avm/server/src/errors.rs
monoid: Actually, I've found that `InterpreterResultDeError` is never constructed. The `AVMInterfaceError` is not needed, one could use `CallSeDeErrors` in the interface.

mikevoronov (approved):
utACK

PR #298, opened 2022-08-30
feat(all): reduce stream usage scope

### LINKED ISSUE #297: Reduce stream usage scope
At the moment streams could be used in different instructions, and some of this cases will lead to undeterministic behaviour (such as #204). And after #292 it's possible prohibit all nondeterministic use cases of streams and replace them with deterministic patterns based on the `canon` instruction.

After fixing this issue stream could be used only:
- in output positions of `ap`/`call` instructions
- as iterable of `fold` instruction without lambda
- as an `argument` of `new` instruction to create an anonymous (in the pi-calculus terms) stream
- as an `argument` of `canon` instruction that allows to canonicalize streams

And all other will be prohibited, including
- using streams as arguments of `call`/`match`/`mismatch` instruction 
- using streams with lambda in triplet positions of `call` instruction and as arguments of `match`/`mismath`/`fail` instructions

Resolves #204.

mikevoronov:
Closes #297.


mikevoronov:
Blocked by #315

mikevoronov:
Blocked by #340

air/src/execution_step/air/mod.rs
```diff
@@ -141,11 +141,11 @@ macro_rules! log_instruction {
 /// This macro converts joinable errors to Ok and sets subgraph complete to false.
 #[macro_export]
 macro_rules! joinable {
-    ($cmd:expr, $exec_ctx:expr) => {
+    ($cmd:expr, $exec_ctx:expr, $ok_result:expr) => {
         match $cmd {
             Err(e) if e.is_joinable() => {
                 $exec_ctx.subgraph_complete = false;
-                return Ok(());
+                return Ok($ok_result);
```
monoid, line 148: It seems all usages of `joinable!` still have `()` as last argument.  It might be useful anyway, but if I have overlooked something, or is it just remnants of some intermediate change?

crates/air-lib/air-parser/src/ast/values.rs
```diff
@@ -23,72 +23,68 @@ use air_lambda_parser::LambdaAST;
...
 #[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
 pub struct Stream<'i> {
     pub name: &'i str,
     pub position: AirPos,
 }
 
-/// A stream with possible lambda expression.
+/// A stream with a lambda expression (it'll be replaced with functors eventually).
 #[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
 pub struct StreamWithLambda<'i> {
```
monoid, line 51: Is not used anymore.

air/src/execution_step/air/mod.rs
mikevoronov: It's just silent "bug" fixing

air/src/execution_step/air/call/triplet.rs
```diff
@@ -44,30 +44,53 @@ pub(crate) fn resolve<'i>(triplet: &ast::Triplet<'i>, ctx: &ExecutionCtx<'i>) ->
 /// Resolve value to string by either resolving variable from `ExecutionCtx`, taking literal value, or etc.
 // TODO: return Rc<String> to avoid excess cloning
```
monoid, line 45: Do these TODO still refer to correct function?  Might they need to be copied to the next one too?

monoid (approved):

monoid:
utACK

air/src/execution_step/air/call/triplet.rs
mikevoronov: Yep, copied, thx

PR #299, opened 2022-09-02
Bump `marine-runtime` to `0.20.0`

nahsi:
Needed for marine e2e to be able to build avm-server using marine-runtime from master.


ValeryAntopol (approved):
I bumped the crate versions, that should be fenough

utACK

PR #307, opened 2022-09-05
Fix fold over canon streams

### LINKED ISSUE #306: Bug: fold fails on an empty canonicalized stream
Using `peek` from the `IterableValue` trait has an assumption that an underlying iterable value is not empty, but this isn't checked for canonicalized streams.

The minimal test case is:
```wasm
(new $stream
    (seq
        (canon "peer_id_1" $stream #canon_stream)
        (fold #canon_stream iterator
            (ap iterator $stream))))
```

It crashes AquaVM being executed.

mikevoronov:
Resolves #306.


ValeryAntopol (approved):
utACK

PR #313, opened 2022-09-06
Fix issue when undefined data in AVM client caused FJS to fail

coder11:


avm/client/src/avmHelpers.ts
```diff
@@ -162,6 +162,9 @@ export async function callAvm(
         return {
             retCode: -1,
             errorMessage: 'marine-js call failed, ' + e,
-        } as any;
+            data: new Uint8Array(),
+            nextPeerPks: [],
+            callRequests: [],
+        };
```
folex, line 168: Now any errors from AVM call will be silently ignored? Or are they reported somewhere else?

coder11: No, fjs will just not fail trying to parse undefined data

avm/client/src/avmHelpers.ts
```diff
@@ -162,6 +162,9 @@ export async function callAvm(
         return {
             retCode: -1,
             errorMessage: 'marine-js call failed, ' + e,
-        } as any;
+            data: new Uint8Array(),
```
mikevoronov, line 165: Can we make a separate result for marine-js, wrapping marine-js into AquaVM error looks wierd.

coder11: Let's just do not handle errors on this level at all. The consumer should be responsible to handle errors in the `fn` passed into arguments

avm/client/src/avmHelpers.ts
coder11: I've removed this block whatsoever. See my answer to Mike

mikevoronov (approved):
utACK

PR #314, opened 2022-09-06
Introduce length functor

### LINKED ISSUE #315: Introduce length functor
#### General

The `length` functor is one of the most simplest and necessary functors for AquaVM. It just returns a size of a variable:
 - for a canonicalized stream it returns count of elements inside
 - for a stream it returns count of elements inside (it a temporal solution and will be removed in #298 !)
 - for an array-like scalar it returns length of this array
 - for a nonarray-like scalar it returns a catchable error

In the text form this functor is similar to lambda, but should be applied without `.$`: 
```wasm
(call "" ("" "") [variable.length])
```

#### Tetraplet algebra

It's possible to create a new variable that will contain length of other variable:

```wasm
(ap scalar.length scalar_length)
```

In this example `scalar_length` will contain a size of `scalar` and it's tetraplet will have all fields empty strings except `json_path` that will contain `.length`.

#### Limitation

For one variable in each place either lambda or functor could be applied. This functor can't be a part of a lambda, so, if you want to apply both lambda and obtain length of its result use `ap`:

```wasm
(seq
  (ap scalar.$.subvalue subvalue)
  (call "" ("" "") [subvalue.length]))
```

mikevoronov:
This PR introduces the `length` functor. To support it `LambdaAST` was completely refactored as well as execution engine parts that handle it. 

#### Breaking change

After this update `.$` part of lambda will be included to the `json_path` field of `SecurityTetraplet`.

Closes #315.


crates/air-lib/lambda/parser/src/lib.rs
```diff
@@ -28,10 +28,11 @@
 mod parser;
 
 pub use parser::parse;
-pub use parser::AccessorsLexer;
-pub use parser::LambdaParser;
+pub use parser::LambdaASTLexer;
+//pub use parser::LambdaParser;
```
monoid, line 32: Stale commented-out code.

air/src/execution_step/boxed_value/jvaluable/cell_vec_resolved_call_result.rs
```diff
@@ -24,29 +24,37 @@ use crate::JValue;
...
+        let select_result = select_by_lambda_from_stream(stream_iter, lambda, exec_ctx)?;
 
-        let tetraplet = &self[select_result.tetraplet_idx].tetraplet;
-        let mut tetraplet = tetraplet.as_ref().clone();
-        tetraplet.add_lambda(&format_ast(lambda));
+        let tetraplet = match select_result.tetraplet_idx {
+            Some(idx) => {
+                let tetraplet = &self[idx].tetraplet;
+                let mut tetraplet = tetraplet.as_ref().clone();
+                tetraplet.add_lambda(&lambda.to_string());
```
monoid, line 53: Can it happen that older tetraplet had `.$.field`, and then we add `.lambda` (functor), getting `.$.field.length`?  I've failed so far to recreate a test for it.

air/src/execution_step/lambda_applier/applier.rs
```diff
@@ -81,14 +122,35 @@ pub(crate) fn select_from_scalar<'value, 'accessor, 'i>(
...
 
-    Ok(value)
+    Ok(Cow::Borrowed(value))
+}
+
+fn select_by_functor_from_scalar(value: &JValue, functor: &Functor) -> ExecutionResult<JValue> {
+    match functor {
+        Functor::Length => {
+            let length = value
+                .as_array()
```
monoid, line 132: Length of a map is also a well-defined operation.

crates/air-lib/interpreter-data/src/interpreter_data.rs
```diff
@@ -54,6 +54,18 @@ pub struct InterpreterData {
     pub restricted_streams: RestrictedStreamGens,
 }
 
+/*
```
monoid, line 57: Leftovers from a call :)

air/src/execution_step/lambda_applier/applier.rs
mikevoronov: Didn't get.

monoid: https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.len

mikevoronov: The reason not to have `length` for object is that `HashMap` for map-type values is not described in a standard, it could be tree of whatever. And imagine that we have this for map-type value why not having it for let's say strings? If so, what about numbers? We could say that length of a number is 1.

And this is the right path to .. JS.

air/src/execution_step/boxed_value/jvaluable/cell_vec_resolved_call_result.rs
mikevoronov: Yep, good catch, I'll rewrite it. I think that smth like

```
(ap scalar.$.field field)
(ap field.length length)
```

should trigger it.

air/src/execution_step/lambda_applier/applier.rs
monoid: Well, having length of string is nice idea. :)

HashMap here is just to give an idea.  Tree has its size too. Yes, we might say that length of a number is 1, but it doesn't mean we should.

Whether we should use same function for vecs and maps is debatable, but if you want separate methods for it, we might rename "length" to something like "vec_length". I'm pretty sure users will need map and string API as well.

air/src/execution_step/boxed_value/jvaluable/cell_vec_resolved_call_result.rs
monoid: Yep, I've managed to get `$.field.length` in the debug output both with a field lambda and field lambda + functor, but I was trying to write a full test initially.

air/src/execution_step/lambda_applier/applier.rs
```diff
@@ -81,14 +122,35 @@ pub(crate) fn select_from_scalar<'value, 'accessor, 'i>(
...
-    pub(self) fn new(result: &'value JValue, tetraplet_idx: u32) -> Self {
+    pub(self) fn from_cow(result: Cow<'value, JValue>, tetraplet_idx: u32) -> Self {
         Self {
             result,
-            tetraplet_idx: tetraplet_idx as usize,
+            tetraplet_idx: Some(tetraplet_idx as usize),
+        }
+    }
+
+    pub(self) fn from_owned(result: JValue) -> Self {
```
monoid, line 150: I believe that "owned" is misleading here.  It's not clear how it differs from `Self::from_cow(Cow::Owned(...))`.

`from_value`?  `from_new_value`? etc.

crates/testing-framework/src/transform/walker.rs
```diff
@@ -57,7 +58,9 @@ impl Transformer {
             self.results.insert(call_id, service.clone());
 
             match &mut call.triplet.1 {
-                Sexp::String(ref mut value) => value.push_str(&format!("..{}", call_id)),
+                Sexp::String(ref mut value) => {
+                    let _ = write!(value, "..{}", call_id);
```
monoid, line 62: Shoudn't it be `.unwrap()`?

CHANGELOG.md
```diff
@@ -1,6 +1,23 @@
...
+[PR 303](https://github.com/fluencelabs/aquavm/pull/303):  
+unsafe code for `LambdaAST` creation was removed
+
+[PR 301](https://github.com/fluencelabs/aquavm/pull/301):  
+bug with adding just executed values to a restricted stream was fixed
+
 ## Version 0.27.0 (2022-08-23)
 
-[PR 292](https://github.com/fluencelabs/aquavm/pull/292):
+[PR 292](https://github.com/fluencelabs/aquavm/pull/292):
```
monoid, line 20: Are these spaces (here and below) really needed?

crates/testing-framework/src/transform/walker.rs
mikevoronov: Hmm, looks like it can't throw an error, but ok, let's add unwrap here

CHANGELOG.md
mikevoronov: Yep, in `.md` file they lead to a newline.

PR #317, opened 2022-09-07
Bump avm-server version to 0.26.0

ValeryAntopol:
The crate version wasn't updated in #299, so this PR will fix it and new version will be finally released.


mikevoronov (approved):
utACK

PR #318, opened 2022-09-08
Use rust-toolchain to setup rust

nahsi:
- Added `rust-toolchain.toml` file so rust toolchain is the same when running rust locally and with CI.
No need to specify channel or components in github workflow file now.

- Changed `Setup rust toolchain` step to use different action. Old one is not maintained anymore. Beware that it sets these env variables by default:
    ```
    CARGO_INCREMENTAL=0
    CARGO_PROFILE_DEV_DEBUG=0
    CARGO_TERM_COLOR=always
    RUST_BACKTRACE=short
    RUSTFLAGS=-D warnings
    ```

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -2,5 +2,5 @@
 # AquaVM can be built with "stable", "nightly" required only to build Marine tests
 channel = "nightly"
 components = [ "rustfmt", "clippy" ]
-targets = [ "x86_64-unknown-linux-gnu", "wasm32-wasi" ]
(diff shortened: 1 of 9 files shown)


rust-toolchain.toml
```diff
@@ -0,0 +1,5 @@
+[toolchain]
+channel = "nightly"
```
mikevoronov, line 2: Let's a comment here that AquaVM itself could be built with `stable`, but `nightly` is required for Marine tests

mikevoronov (approved):
utACK

PR #320, opened 2022-09-08
AVM client: Replace `any` with generic JSON types

coder11:


ValeryAntopol (approved):
utACK

Please bump some versions if needed. I am not sure if any are required, so its up to you.

mikevoronov (approved):
utACK

PR #324, opened 2022-09-12
ci: Add Aquavm E2E

nahsi:


.github/workflows/e2e.yml
```diff
@@ -0,0 +1,127 @@
...
+        with:
+          context: .
+          push: true
+          file: .github/snapshot/Dockerfile
+          tags: |
+            docker.fluence.dev/aquavm:${{ github.head_ref }}
+          labels: |
+            sha=${{ github.event.pull_request.head.sha }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
```
folex, line 115: Why not move it to `snapshot.yml` in rust peer repo and call it like
```
  rust-peert:
     uses: fluencelabs/rust-peer/.github/workflows/snapshot.yml@main
     with:
       rust-peer-foo-bar: ...
```

nahsi: Yes, that is a good idea, maybe in another PR?

.github/workflows/snapshot.yml
```diff
@@ -0,0 +1,153 @@
...
+      - name: Checkout Aquavm
+        uses: actions/checkout@v3
+
+      - name: Setup Rust toolchain
+        uses: actions-rust-lang/setup-rust-toolchain@v1
+
+      - name: Install marine
+        uses: baptiste0928/cargo-install@v1
+        with:
+          crate: marine
```
folex, line 32: maybe use `download_marine.sh`? it's usually faster by an order or two

avm/client/build_wasm.sh
```diff
@@ -1,9 +1,13 @@
-#!/bin/sh
-(
-  cd ../../air-interpreter;
-  marine build  --release --features marine
-)
+#! /usr/bin/env bash
 
-mkdir -p ./dist/
-cp ../../target/wasm32-wasi/release/air_interpreter_server.wasm ./dist/avm.wasm
+if [[ -f dist/avm.wasm ]]; then
```
folex, line 3: That would prevent rebuilding after `cargo clean`, afaiu. Because `cargo clean` won't remove `dist/avm.wasm`.

I'm not sure it is a problem, cuz I'm not sure how `avm/client/build_wasm.sh` is used by humans, if at all.

.github/workflows/snapshot.yml
nahsi: image :)

PR #325, opened 2022-09-12
wasm-air-beautify WASM module

monoid:
It provides a `beautify` method to output a "beautified" (python-style formatted) AIR script.

The method returns an output string or raises an exception.


mikevoronov (approved):
utACK

PR #332, opened 2022-09-14
Fix canon positions

### LINKED ISSUE #331: Bug: ap with canon uses incorrect position
The ap instruction sets an incorrect position for a resulting value.

The minimal example:

```wasm
(new $array-inline
    (seq
        (seq
            (seq
                (seq
                    (call %init_peer_id% ("op" "array_length") [$status] array_length)
                    (ap array_length $array-inline)
                )
                (seq
                    (ap 2 $num)
                    (canon %init_peer_id% $num #num_canon)
                )
            )
            (ap #num_canon.$.[0]! $array-inline)
        )
        (canon %init_peer_id% $array-inline #array-inline-0)
   )
)
```

mikevoronov:
This PR changes scheme how positions in `ap` are assigned for values constructed from canonicalized streams.

Resolves #331.


ValeryAntopol (approved):
utACK

PR #335, opened 2022-09-19
Introduce `never` instruction

### LINKED ISSUE #334: Introduce instruction `never`
There is instruction `null` that simply does nothing and doesn't influence on subgraph completeness. This instruction was introduced for code generation purposes. And sometimes it's needed to have another instruction that also does nothing, but set the completeness of a subgraph to false.

mikevoronov:
Closes #334.


monoid:
Add it to the README, please.

monoid (approved):

monoid:
It worth incrementing the air-parser version too, as there are other packages that depend on its change.

mikevoronov:
Not in this PR since the internal hackathon is coming

PR #337, opened 2022-09-20
fix(execution-engine) Restricted stream bugfix

### LINKED ISSUE #302: Bug: values from current data aren't set properly to a restricted stream
AIR streams have global or scoped visibility, also streams have generations and each new AquaVM launch all new values should be placed into a new generation. There were two problems with that for restricted streams: #300 was about adding just executed call result to a proper generation and this issue is about adding value from current data. The crux of the problem is that TraceHandler carefully choose a proper generation by using `GlobalStreamGens` variable, but it doesn't contain `RestrictedStreamGens` for restricted streams and so doesn't change generations in the `merge_current_executed` function.

The minimal AIR example is 

```wasm
(new $stream
  (par
    (call "peer_id_1" ("" "") [2] $stream)
    (seq
      (call "peer_id_2" ("" "") [1] $stream)
      (seq
        (call "peer_id_3" ("" "") [0] $stream)
        (call "peer_id_2" ("" "") [$stream])))))
```

Being run on `peer_vm_2` -> `peer_vm_1` -> `peer_vm_3` -> `peer_vm_2`, the expected trace will be:
```
[
  0: par(1, 3)
  1: executed(stream: 2 generation: 1)
  2: executed(stream: 1 generation: 0)
  3: executed(stream: 0 generation: 1)
  4: executed(scalar: [1,2,0])
]
```

But the actual is:
```
[
  0: par(1, 3)
  1: executed(stream: 2 generation: 0)
  2: executed(stream: 1 generation: 0)
  3: executed(stream: 0 generation: 0)
  4: executed(scalar: [2,1,0])
]
```

And the actual trace should be equal to AIR when `$stream` is global.

monoid:
Closes #302.


air/src/execution_step/air/call/prev_result_handler.rs
```diff
@@ -84,8 +86,9 @@ pub(super) fn handle_prev_state<'i>(
             Ok(StateDescriptor::cant_execute_now(prev_result))
         }
         // this instruction's been already executed
-        Executed(value) => {
-            set_result_from_value(value.clone(), tetraplet.clone(), trace_pos, output, exec_ctx)?;
+        Executed(ref mut value) => {
+            eprintln!("Scheme: {:?}, {:?}", scheme, value);
```
mikevoronov, line 90: debug print

mikevoronov (approved):
utACK

PR #340, opened 2022-09-26
feat(execution engine): change behaviour of fold over streams

### LINKED ISSUE #333: Change behaviour of fold over stream
At the moment it has the following behaviour:
 - if stream is empty then fold does nothing and doesn't mark subgraph as incomplete
 - if stream is non empty then fold iterates over values and if either par/xor/seq combinators are used fold execution could be finished, i.e. subgraph would be completed

And the proposed changes are
 - if stream is empty then a fold marks subgraph as incomplete (it would be like the join behaviour)
 - if next is met in a fold over stream it also marks a subgraph as incomplete. It's the breaking change and has some important consequences:
    - fold with a seq combinator and "traditional" usage of next can't finish (i.e. set subgraph to complete), so the following
    (fold $stream iterator
         (seq
          ...
          (next iterator)))
    will never finished. We could consider (next iterator) as a listen on an infinite stream
    - fold with a par combinator will finish if any it's subtree finished, f.e.
    (fold $stream iterator
       (par
          (call "" ("service_name" "") [])
          (next iterator)))
    finished if at least one call is succeeded
    - fold with a xor completely depends on a place of next
    
**Connection with pi-calculus**

Pi-calculus is highly relies on a concept of streams/names/channels/... that aims to send and receive values from. A part of expression that uses stream for reciive value from is typically waited on it and assume that streams are infinite. F.e.
`x!5 | x?a`
this means that there are two part of a statement run in parallel, and the left sends 5 to a channel named x, whether the right "waits" on this channel and receives a value denoted by a.
After the proposed change our streams would have almost the same algebra.

mikevoronov:
Resolves #333.


air/tests/test_module/features/streams/recursive_streams.rs
```diff
@@ -333,43 +339,21 @@ fn recursive_stream_inner_fold() {
...
+                            (next iterator_1))))))
+        (call "{vm_peer_id_2}" ("" "") ["{result_value}"]))
+       "#);
 
     let result = checked_call_vm!(vm_1, <_>::default(), &script, "", "");
     let result = checked_call_vm!(vm_2, <_>::default(), script, "", result.data);
     let actual_trace = trace_from_result(&result);
 
-    let actual_last_state = &actual_trace[22.into()];
+    let actual_last_state = &actual_trace[(actual_trace.len() - 1).into()];
```
monoid, line 354: Можно `.last().unwrap()`, но это мелочь.

air/src/execution_step/air/fold_stream.rs
```diff
@@ -96,22 +102,10 @@ fn execute_iterations<'i>(
             trace_ctx,
         );
         trace_to_exec_err!(trace_ctx.meet_generation_end(fold_id), fold_stream)?;
-
-        result?;
-        if !exec_ctx.subgraph_complete {
-            break;
-        }
+        generation_observer.observe_generation_results(exec_ctx.subgraph_complete, result);
```
monoid, line 105: I wonder why don't we stop at the first error, but return the last one.

mikevoronov: It's the previous behaviour and after the latest changes it became not deterministic. The issue here in gates, to make them all open deterministically it's crucial to go through all executed states.

monoid (approved):

PR #341, opened 2022-09-27
refactor(parser): use `AirPos` type for AIR script position

monoid:
Use a dedicated wrapper type for better type safety and self-documented code.


mikevoronov:
utACK.

Please follow the conventional commit scheme of commit messages.

mikevoronov (approved):

PR #342, opened 2022-09-27
feat(testing) Testing framework chapter 1, asserts and comments

monoid:
1. Rename `seq_result` -> `seq_ok`, and add `seq_err`.
2. Accept `;;` and longer comments in the sexp parser.


mikevoronov (approved):
utACK

crates/testing-framework/src/execution/mod.rs
```diff
@@ -322,6 +325,137 @@ mod tests {
...
+                    executed_state::scalar(json!(["peer2", "peer3"])),
+                    executed_state::scalar(json!(42)),
+                    executed_state::scalar(json!(43)),
+                ]
+            );
+        }
+    }
+
+    #[test]
+    fn test_seq_error() {
```
ValeryAntopol, line 384: This test looks too large. Maybe split is in several ones?

crates/testing-framework/src/services/results.rs
```diff
@@ -74,6 +77,37 @@ fn seq_result_closure(call_map: HashMap<String, serde_json::Value>) -> CallServi
...
+                    call_num_str, call_map
+                )
+            })
+            .clone()
+    })
+}
+
+fn map_service_closure(map: HashMap<String, serde_json::Value>) -> CallServiceClosure {
+    Box::new(move |args| {
+        let repr = match &args.arguments[0] {
```
ValeryAntopol, line 103: What if args are empty? I suggest panic with a good error message.

ValeryAntopol (approved):
utACK

I think comments fixes do not require additional review

crates/testing-framework/src/execution/mod.rs
monoid: It going to be several long tests.

PR #347, opened 2022-09-29
fix(execution-engine): treat non-defined stream as empty in `canon`

### LINKED ISSUE #346: Bug: `canon` for not-defined stream leads to non-deterministic behaviour
The following script

```wasm
(par
    (call "unknown_peer" ("" "") [] $stream) ; to make validator happy
    (xor
        (canon "vm_peer_id" $stream #canon_stream) ; it returns a catchable error, because stream hasn't been defined
        (call "vm_peer_id" ("" "") [""])
    )
)
```

fails because `canon` throws an error if stream wasn't defined.

From the second glance it could lead to non-deterministic behavior, because stream could be populated later and so `canon` won't throw an error.

mikevoronov:
This PR forces `canon` instruction to treat streams as empty, it's crucial for deterministic behaviour.

Closes #346.

### DIFF
--- a/air/src/execution_step/air/canon.rs
+++ b/air/src/execution_step/air/canon.rs
@@ -19,9 +19,9 @@ use super::ExecutionResult;
 use super::TraceHandler;
 use crate::execution_step::boxed_value::CanonStream;
 use crate::execution_step::Generation;
+use crate::execution_step::Stream;
 use crate::log_instruction;
 use crate::trace_to_exec_err;
-use crate::CatchableError;
 use crate::ExecutionError;
 use crate::UncatchableError;
 
@@ -30,7 +30,7 @@ use air_interpreter_data::TracePos;
 use air_parser::ast;
 use air_trace_handler::MergerCanonResult;
 
-use std::rc::Rc;
+use std::borrow::Cow;
 
 impl<'i> super::ExecutableInstruction<'i> for ast::Canon<'i> {
     #[tracing::instrument(level = "debug", skip(exec_ctx, trace_ctx))]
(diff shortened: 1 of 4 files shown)


air/src/execution_step/air/canon.rs
```diff
@@ -169,3 +155,17 @@ fn create_canon_stream_from_name(
...
+
+/// This function gets a stream from context or return a default empty stream,
+/// it's crucial for deterministic behaviour, for more info see
+/// github.com/fluencelabs/aquavm/issues/346
+fn get_stream_or_default<'ctx, 'value>(
+    ast_canon: &ast::Canon<'_>,
+    exec_ctx: &'ctx ExecutionCtx<'value>,
+) -> Cow<'ctx, Stream> {
+    let maybe_stream = exec_ctx.streams.get(ast_canon.stream.name, ast_canon.stream.position);
+    match maybe_stream {
```
monoid, line 167: YMMV, but it seems to be `maybe_stream.map(Cow::Borrowed).unwrap_or_default()`.

air/src/execution_step/air/canon.rs
```diff
@@ -169,3 +155,17 @@ fn create_canon_stream_from_name(
...
 }
+
+/// This function gets a stream from context or return a default empty stream,
+/// it's crucial for deterministic behaviour, for more info see
+/// github.com/fluencelabs/aquavm/issues/346
+fn get_stream_or_default<'ctx, 'value>(
+    ast_canon: &ast::Canon<'_>,
+    exec_ctx: &'ctx ExecutionCtx<'value>,
+) -> Cow<'ctx, Stream> {
+    let maybe_stream = exec_ctx.streams.get(ast_canon.stream.name, ast_canon.stream.position);
```
monoid, line 166: Do you need to pas whole `exec_ctx`, or `streams` is enough?

air/src/execution_step/air/canon.rs
mikevoronov: yep, it works, thank you

monoid (approved):

monoid:
utACK

air/src/execution_step/air/canon.rs
mikevoronov: `Streams` is enough, yeah, but it would require importing it into this file

PR #350, opened 2022-09-30
fix(trace-handler): treat `ap` states with different generations as compatible for `canon`

### LINKED ISSUE #348: Bug: canon misnterprets some states in the trace handler as incompatible
### Example:
```
        (seq
            (seq
                (ap 1 $inner)
                (seq
                    (call "{vm_peer_id_2}" ("op" "noop") [1])
                    (ap 2 $inner)
                )
            )
            (seq
                (canon "{vm_peer_id_2}" $inner #inner)
                (seq
                    (call "{vm_peer_id_3}" ("op" "noop") [1])
                    (call "{vm_peer_id_2}" ("op" "noop") [1])
                )
            )
        )
```
WIth the call order `vm1->vm2->vm3->vm2`. On the last call vm2 should use second call's data as prev_data.

### Error text:
```VM returns a error: 20000 on instruction 'canon "vm_peer_id_2" $inner #inner' trace handler encountered an error: canon results CanonResult { stream_elements_pos: [0, 2] } CanonResult { stream_elements_pos: [0, 2] } at position 1 points to incompatible execution states: Some(Ap(ApResult { res_generations: [1] })) Some(Ap(ApResult { res_generations: [0] }))```

### Explaination:
On the second call instruction `(ap 2 $inner)` is first met, and it creates a new generation for stream `$inner`, and produces corresponding state in trace: `ApResult { res_generations: [1] })`, which is expected. On the third call, second `ap` instruction does not create a new generation and puts `ApResult { res_generations: [0] })`, which is also expected. The error happens on the last call on `vm2`, where stored prev_data and current_data are merged. `canon` state consists of list of trace positions. During merege they are checked for compatibility. Merger compares also the content of the `ap` states, which is invalid. So, the fix is to remove this check.

ValeryAntopol:
Fixes #348

### DIFF
--- a/air/tests/test_module/issues/issue_348.rs
+++ b/air/tests/test_module/issues/issue_348.rs
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2022 Fluence Labs Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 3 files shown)


mikevoronov (approved):
utACK

PR #351, opened 2022-10-01
feat(execution-engine)!: make fold convergent wrt errors

mikevoronov:
`fold` over stream is not convergent when errors are produced inside a fold body. After iteration it produces if there were several errors only the last one is bubbled up. But on the same peer there could different last (and first) errors and it'll became non-convergent and moreover non-deterministic.

After this PR `fold` won't bubble any errors to make execution convergent and deterministic. To obtain errors from a fold body one should wrap this body into `xor` and push errors into some stream. Then fold over this stream and handle errors.


air/tests/test_module/features/streams/streams_early_exit.rs
```diff
@@ -178,59 +178,35 @@ fn par_early_exit() {
 }
 
 #[test]
+// TODO: rewrite this test
```
monoid, line 181: How?

mikevoronov: It's an obsolate

air/src/execution_step/air/fold_stream.rs
```diff
@@ -128,3 +130,13 @@ fn remove_new_generation_if_non_empty<'ctx>(
     stream.remove_last_generation_if_empty();
     stream
 }
+
+/// Fold over streams doesn't throw an error if it not catchable, because otherwise it would be
```
monoid, line 134: I believe one negation in this line should be removed.

monoid:
utACK

monoid (approved):

PR #353, opened 2022-10-06
refactor(aquavm): getting rid of `CallOutputValue` in call merger

mikevoronov:
This PR refactors call merger of `TraceHandler`. Previously it requires `CallOutputValue` to determine a type of call output value (stream or scalar). And internally it checked correspondence betweendata result and call output type and return a error if they are not equal. Although execution engine component also had a `match` over these values and does nothing if they are not matched since `TraceHandler` did this job. This PR eliminate such behaviour and improve isolation of AquaVM modules.

### DIFF
--- a/air/src/execution_step/air/call/prev_result_handler.rs
+++ b/air/src/execution_step/air/call/prev_result_handler.rs
@@ -15,7 +15,7 @@
  */
 
 use super::*;
-use crate::execution_step::air::call::call_result_setter::set_result_from_value;
+use crate::execution_step::air::call::call_result_setter::populate_context_from_data;
 use crate::execution_step::CatchableError;
 use crate::execution_step::RcSecurityTetraplet;
 
@@ -84,7 +84,7 @@ pub(super) fn handle_prev_state<'i>(
         }
         // this instruction's been already executed
         Executed(value) => {
-            let resulted_value = set_result_from_value(
+            let resulted_value = populate_context_from_data(
                 value.clone(),
                 tetraplet.clone(),
                 met_result.trace_pos,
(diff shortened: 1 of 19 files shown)


air/src/execution_step/air/call/prev_result_handler.rs
```diff
@@ -71,24 +68,32 @@ pub(super) fn handle_prev_state<'i>(
...
-            Ok(StateDescriptor::cant_execute_now(prev_result))
+            Ok(StateDescriptor::cant_execute_now(met_result.result))
         }
         // this instruction's been already executed
-        Executed(ref mut value) => {
-            set_result_from_value(&mut *value, tetraplet.clone(), trace_pos, scheme, output, exec_ctx)?;
-            trace_ctx.meet_call_end(prev_result);
+        Executed(value) => {
+            let resulted_value = populate_context_from_data(
+                value.clone(),
```
monoid, line 88: This clone is cheap, but one can `match met_result.result` and avoid it.

crates/air-lib/trace-handler/src/merger/call_merger.rs
```diff
@@ -18,66 +18,57 @@ mod utils;
...
-    output_value: &CallOutputValue<'_>,
-) -> MergeResult<MergerCallResult> {
+#[derive(Debug, Clone)]
+pub struct MetResult {
+    pub result: CallResult,
+    pub trace_pos: TracePos,
+    pub source: ValueSource,
+}
+
+#[derive(Debug, Clone)]
```
monoid, line 42: I would either add `Copy` here, or remove `Clone` here and above (it compiles!).

air/src/execution_step/air/call/prev_result_handler.rs
mikevoronov: Good catch, thanks

ValeryAntopol (commented):
utACK

PR #354, opened 2022-10-07
refactor(aquavm): improve readability of `ap` merger and handler

### LINKED ISSUE #355: Refactor ap state and merger
At the moment `Ap` state contains vector of generations, although the `ap` instruction itself allow to specify only one stream. It was done to support several streams in a result position in a future. But the current realization allows to handle empty vector that was implemented to support states for scalars and it turned out (for more info see https://github.com/fluencelabs/aquavm/issues/326#issuecomment-1261326472) that states for scalars aren't actually needed.

mikevoronov:
Closes #355.

Contains improvements for bug fixing in #356.


crates/air-lib/test-utils/src/executed_state.rs
```diff
@@ -132,21 +132,12 @@ pub fn subtrace_desc(begin_pos: impl Into<TracePos>, subtrace_len: u32) -> SubTr
     }
 }
 
-pub fn ap(dst: Option<u32>) -> ExecutedState {
-    let res_generations = option_to_vec(dst);
-    let ap_result = ApResult::new(res_generations);
-
+pub fn ap(generation: u32) -> ExecutedState {
+    let ap_result = ApResult::new(vec![generation]);
```
monoid, line 136: The `ApResult::new` can be further refactored to accept a single number: 4136257804228ba6e69f626739c5e6917f3ad0c9

monoid (approved):

air/src/execution_step/air/ap/utils.rs
```diff
@@ -17,56 +17,31 @@
...
-    use crate::execution_step::UncatchableError::ApResultNotCorrespondToInstr;
-    use ast::ApResult::*;
-
-    match (variable, generation) {
-        (Stream(_), Some(_)) => Ok(()),
-        (Scalar(_), None) => Ok(()),
-        _ => Err(ApResultNotCorrespondToInstr(ap_result.clone()).into()),
+    match met_result.value_source {
+        ValueSource::PreviousData => Generation::Nth(met_result.generation),
+        ValueSource::CurrentData => Generation::Last,
```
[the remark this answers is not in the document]
mikevoronov: Yep, we had a lot of problems with `_` previously, that why I want to avoid this variant.

crates/air-lib/test-utils/src/executed_state.rs
mikevoronov: Yep, will change it.

PR #358, opened 2022-10-10
fix(aquavm): temporary fix entire value in canon

mikevoronov:


crates/air-lib/interpreter-data/src/executed_state/impls.rs
```diff
@@ -129,8 +129,8 @@ impl std::fmt::Display for ExecutedState {
             Ap(ap) => {
                 write!(f, "ap: _ -> {:?}", ap.res_generations)
             }
-            Canon(canon) => {
-                write!(f, "canon {:?}", canon.stream_elements_pos)
+            Canon(_) => {
+                write!(f, "canon [<object>]")
```
monoid, line 133: We loose information.

crates/air-lib/trace-handler/src/merger/canon_merger.rs
```diff
@@ -58,75 +51,33 @@ pub(crate) fn try_merge_next_state_as_canon(data_keeper: &mut DataKeeper) -> Mer
...
     })
 }
 
 fn check_canon_results(
     prev_canon_result: &CanonResult,
     current_canon_result: &CanonResult,
-    data_keeper: &DataKeeper,
 ) -> Result<(), CanonResultError> {
-    if prev_canon_result.stream_elements_pos.len() != current_canon_result.stream_elements_pos.len() {
+    if prev_canon_result.canonicalized_element.len() != current_canon_result.canonicalized_element.len() {
```
monoid, line 69: We might compare content, if you care.

monoid (approved):

crates/air-lib/test-utils/src/executed_state.rs
```diff
@@ -137,7 +137,12 @@ pub fn ap(generation: u32) -> ExecutedState {
     ExecutedState::Ap(ap_result)
 }
 
-pub fn canon(canonicalized_element: Vec<u8>) -> ExecutedState {
+pub fn canon(_canonicalized_element: Vec<u8>) -> ExecutedState {
```
monoid, line 140: Variable is unused. Is it valid? Is the function used?

mikevoronov: Thanks, forgot to delete.

PR #357, opened 2022-10-08
fix(trace-handler): fix fold and canon compatibility

mikevoronov:
Closes #356.

PR #362, opened 2022-10-10
fix(execution-engine): fix invalid iteration over stream

### LINKED ISSUE #363: Bug(execution engine): interpreter loses call results from current_data
Air:
```
        (seq
            (seq
                (call "client" ("get" "data") [] permutations) ; ok = [["p1",[[["p1",1],["p3",3]],[["p1",4],["p3",5]]]],["p3",[[["p3",13],["p1",14]],[["p3",16],["p1",18]]]]]
                (seq
                    (fold permutations pair
                        (seq
                            (fold pair.$.[1] peer_ids
                                (seq
                                    (seq
                                        (call pair.$.[0] ("op" "noop") []) ; ok = null
                                        (ap peer_ids $inner)
                                    )
                                    (next peer_ids)
                                )
                            )
                            (next pair)
                        )
                    )
                    (seq
                        (null)
                        (fold $inner ns
                            (par
                                (fold ns pair
                                    (seq
                                        (seq
                                            (call pair.$.[0] ("op" "noop") []) ; ok = null
                                            (ap pair.$.[1] $result)
                                        )
                                        (next pair)
                                    )
                                )
                                (next ns)
                            )
                        )
                    )
                )
            ) (null)
        )
```

Execution flow: `client -> p1 -> p3 -> p1`
Value for variable `permutations` is in comment for the corresponding call.

Error summary:
`(fold $inner ns` executes inner fold on 4 groups of pairs, two groups request execution first on p3, then on p1, another two request execution first on p1, then on p3. So, on the p3 call two groups are partially executed, then results sent to p1.  This peer is expected to retain the results from p3, execute calls to services on p1 and then pass the result back to p3. But instead it loses results from p3, just executes its own calls.

mikevoronov:
This PR is mostly a revertion of https://github.com/fluencelabs/aquavm/pull/357, that is needed to make stream work correctly in fold itrerations.

Closes #363.


CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## Version 0.31.0 (2022-10-11)
+
+[PR 360](https://github.com/fluencelabs/aquavm/pull/360):
```
ValeryAntopol, line 3: Add this PR also

air/Cargo.toml
```diff
@@ -41,7 +41,7 @@ strum_macros = "0.24"
 tracing = "0.1.36"
 
 [dev_dependencies]
-air-test-utils = { path = "../crates/air-lib/test-utils" }
+air-test-utils = { path = "../crates/air-lib/test-utils",  features = ["test_with_native_code"] }
```
ValeryAntopol, line 44: Remove this?

ValeryAntopol (approved):
tested ACK

PR #364, opened 2022-10-11
feat(testing-framework): allow exclamation mark

monoid:
While exclamation mark is legacy feature in the lambda parser, people seem to use it anyway, getting errors not always easy to understand.


mikevoronov (approved):
utACK

PR #365, opened 2022-10-11
feat(testing-framework): prevalidate AIR script

monoid:
Try to parse the annotated AIR script with the standard parser to catch to catch errors early with better error messages.

We do it in the `TestExecutor` only to make testing framwork parser tests simplier.  For user experience, it doesn't matter.


mikevoronov (approved):
utACK

PR #367, opened 2022-10-13
feat(interpreter-data): add interpreter version in data

mikevoronov:


air/tests/test_module/features/misc/version_check.rs
```diff
@@ -0,0 +1,35 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+use air::PreparationError;
+use air_test_utils::prelude::*;
+
+#[test]
```
folex, line 20: i would add a test for our branch versions. it seems this will break them, cuz they're usually 0.0.1-foo-bar

mikevoronov: Good catch

folex (approved):

air/src/preparation_step/minimal_supported_version.rs
```diff
@@ -0,0 +1,30 @@
...
+
+use once_cell::sync::Lazy;
+
+use std::str::FromStr;
+
+static MINIMAL_SUPPORTED_VERSION: Lazy<semver::Version> =
+    Lazy::new(|| semver::Version::from_str("0.31.2").expect("valid minimal supported version specified"));
+
+// This local is intended to check that set version is correct at the AquaVM start for graceful error message.
+thread_local!(static MINIMAL_SUPPORTED_VERSION_CHECK: &'static semver::Version = Lazy::force(&MINIMAL_SUPPORTED_VERSION));
```
monoid, line 25: It seems to be not used by name.  Let's precede it with an underscore.

air/src/preparation_step/errors.rs
```diff
@@ -33,12 +33,19 @@ pub enum PreparationError {
 
     /// Errors occurred on executed trace deserialization.
     #[error("an error occurred while executed trace deserialization on {1:?}:\n{0:?}.\
-    Probably it's a data of an old version that couldn't be converted to '{}'", *DATA_FORMAT_VERSION)]
+    Probably it's a data of an old version that can't be converted to '{}'", *data_version())]
```
monoid, line 36: It's minor, but I think the dereference is not needed here.

coder11 (approved):

monoid (approved):

PR #375, opened 2022-11-02
feat(testing-framework): testing parser improvements

monoid:
1. Multiline annotations with `#| ... |#` syntax.  Useful for large JSON values.
2. Assertions after nested statements (refers to rightmost nested form).
3. JSON parser, so that multiple fields with JSON values can be used in the future.


mikevoronov (approved):
utACK

PR #381, opened 2022-11-18
feat!(avm-interface): Allow overriding current_peer_id

monoid:
The current peer ID is passed as a new field of `TestRunParameters` named
`current_peer_id: String`, instead of creating an AVM with peer ID.

This is a breaking API change of `avm-interface` and `avm-server`.


folex (approved):
It seems from the code that we might rename `override_current_peer_id: Option<String>` into `current_peer_id: String`, so that it will be passed all the time.

avm/interface/src/particle_parameters.rs
```diff
@@ -20,25 +20,30 @@ use std::borrow::Cow;
 
 /// Represents parameters obtained from a particle.
 #[derive(Debug, Clone, Serialize, Deserialize)]
-pub struct ParticleParameters<'init_peer_id, 'particle_id> {
+pub struct ParticleParameters<'init_peer_id, 'particle_id, 'current_peer_id> {
```
monoid, line 23: @mikevoronov I think we should use single lifetime here.

mikevoronov: Agree

mikevoronov (approved):
utACK

air/tests/test_module/integration/override_current.rs
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+use air_test_utils::prelude::*;
+use fstrings::f;
+use fstrings::format_args_f;
+
+use std::collections::HashMap;
+
+#[test]
+fn test_override_current_peer_id() {
```
monoid, line 24: The test now tests nothing but test-utils and should be moved elsewhere.

justprosh (approved):

PR #385, opened 2022-11-23
ci: Renovate tuning

nahsi:
- do not group versions update in one PR (unless grouped by default i.e. `jest`)
- max 10 open PR
- max 1 open major bump PR
- dev dependencies have lowest priority
- runs only on weekends


mikevoronov (approved):
utACK

PR #396, opened 2022-11-28
feat(testing-framework): tetraplet behavior

monoid:
Add `behavior=tetraplet` for testing framework annotation to return call's tetraplet.

### DIFF
--- a/crates/testing-framework/src/services/results.rs
+++ b/crates/testing-framework/src/services/results.rs
@@ -19,7 +19,7 @@ use crate::asserts::ServiceDefinition;
 
 use air_test_utils::CallRequestParams;
 
-use std::{cell::RefCell, collections::HashMap, rc::Rc, time::Duration};
+use std::{cell::RefCell, collections::HashMap, rc::Rc};
 
 #[derive(Clone, Debug, Default, PartialEq, Eq)]
 pub(crate) struct ResultStore {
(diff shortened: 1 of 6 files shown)


crates/testing-framework/src/services/results.rs
```diff
@@ -36,13 +36,16 @@ impl ResultStore {
...
+    fn call(&self, mut params: CallRequestParams) -> FunctionOutcome {
         let results = self.results.borrow();
-        if let Some((_, suffix)) = params.service_id.split_once("..") {
-            if let Ok(key) = suffix.parse() {
-                let service_desc = results.get(&key).expect("Unknown result id");
-                let service_result = service_desc.call(params);
-                FunctionOutcome::ServiceResult(service_result, Duration::ZERO)
+        if let Some((real_service_id, suffix)) = params.service_id.rsplit_once("..") {
+            if let Ok(result_id) = suffix.parse() {
+                let service_desc = results.get(&result_id).unwrap_or_else(|| {
```
mikevoronov, line 43: let's decrease identation level here with `match`

mikevoronov (approved):
utACK

crates/testing-framework/src/services/results.rs
monoid: Like this?

mikevoronov: Yep, thanks

PR #402, opened 2022-12-12
Make unstable clippy happy with `clippy --fix`

monoid:
Most of changes either move variables into `format!` templates or remove excessive clones.


mikevoronov (approved):
utACK

PR #411, opened 2022-12-26
fix(interpreter-cid): Fix interpreter-cid compilation

monoid:
The serde crate was missing "derive" feature.


mikevoronov (approved):
utACK

PR #412, opened 2022-12-26
fix(interpreter-cid): fix Cargo.toml keyword list

monoid:
It was too long, and the crate was rejected by the crates.io.


mikevoronov (approved):
utACK

PR #414, opened 2022-12-26
fix(interpreter-interface): Remove the `cid` field

monoid:
The `InterpreterOutcome::cid` field is not really needed, but makes integration more difficult.


mikevoronov (approved):
utACK,
don't forget to update the main CHANGELOG

PR #416, opened 2022-12-26
chore(execution-engine): more handy way to work with subgraph_complete

mikevoronov:


air/src/execution_step/execution_context/context.rs
```diff
@@ -115,6 +115,24 @@ impl<'i> ExecutionCtx<'i> {
...
+impl ExecutionCtx<'_> {
+    pub(crate) fn make_subgraph_incomplete(&mut self) {
+        self.subgraph_completeness = false;
+    }
+
+    pub(crate) fn is_subgraph_complete(&self) -> bool {
+        self.subgraph_completeness
+    }
+
+    pub(crate) fn set_subgraph_completeness(&mut self, subgraph_complete: bool) {
```
monoid, line 127: I think `set_subgraph_complete` sounds better (and safe for the next method).

mikevoronov: Maybe `make_subgraph_complete`?

monoid: Well, it doesn't make it always complete.
Let's stick with the original name then.

monoid (approved):

PR #429, opened 2023-01-10
fix(air-trace): `air-trace` native mode compilation fix

monoid:
It was not correctly refactored in the #381.


mikevoronov (approved):
utACK

mikevoronov (approved):
utACK

PR #451, opened 2023-02-06
feat(trace-handler): improve data deserialization version check

mikevoronov:
Aims to make an error message more clear when data of incompatible versions is used.

### DIFF
--- a/air/src/preparation_step/errors.rs
+++ b/air/src/preparation_step/errors.rs
@@ -16,6 +16,7 @@
 
 use crate::ToErrorCode;
 use air_interpreter_data::data_version;
+use air_interpreter_data::Versions;
 
 use serde_json::Error as SerdeJsonError;
 use strum::IntoEnumIterator;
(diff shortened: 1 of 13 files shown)


air/src/preparation_step/errors.rs
```diff
@@ -33,9 +33,10 @@ pub enum PreparationError {
 
     /// Errors occurred on executed trace deserialization.
     #[error(
-        "an error occurred while executed trace deserialization: {0:?}.\
-    Probably it's a data of an old version that can't be converted to '{}'.\n\
-    Data: {1:?}",
+        "an error occurred while executed trace deserialization: {0:?}.\n\
+        Interpreter of {} version is used and it's expecting a data of {} version.\n\
+        Data: {1:?}",
+        super::interpreter_version(),
```
folex, line 39: wdyt about putting `interpreter_version` into data? So that we can say "got data from interpreter X.X.X, support minimum Y.Y.Y, current is Z.Z.Z"?

mikevoronov: This error was split into two: the first when versions weren't possible to recover from data and the second when it was possible.

air/src/preparation_step/errors.rs
```diff
@@ -33,19 +34,42 @@ pub enum PreparationError {
 
     /// Errors occurred on executed trace deserialization.
     #[error(
-        "an error occurred while executed trace deserialization: {0:?}.\
-    Probably it's a data of an old version that can't be converted to '{}'.\n\
-    Data: {1:?}",
+        "an error occurred while data deserialization: {error:?}.\n\
+        Your AquaVM's version is {} and it's expecting data of {} version,\
```
folex, line 38:
```suggestion
        AquaVM's version is {} and it expects data of {} version,\
```

folex (approved):
awesome!

air/src/preparation_step/preparation.rs
```diff
@@ -75,10 +75,20 @@ pub(crate) fn prepare<'i>(
...
+    }
+
+    InterpreterData::try_from_slice(raw_data).map_err(|de_error| to_date_de_error(raw_data.to_vec(), de_error))
+}
 
-    InterpreterData::try_from_slice(raw_data, super::min_supported_version())
-        .map_err(|err| DataDeFailed(err, raw_data.to_vec()))
+fn to_date_de_error(raw_data: Vec<u8>, de_error: serde_json::Error) -> PreparationError {
+    match InterpreterData::try_get_versions(&raw_data) {
+        Ok(versions) => PreparationError::data_de_failed_with_versions(raw_data, de_error, versions),
```
monoid, line 89: Is there a test for this path?

mikevoronov: added

monoid (approved):

PR #492, opened 2023-03-01
feat(execution-engine)!: make StreamDontHaveSuchGeneration uncatchable…

mikevoronov:
StreamDontHaveSuchGeneration is a catchable error that could be caught by a xor instruction and then handled by a user. But it makes no sense because this error could arise only iff smth went wrong during merging or when data is corrupted.

### DIFF
--- a/air/src/execution_step/boxed_value/stream.rs
+++ b/air/src/execution_step/boxed_value/stream.rs
@@ -88,7 +88,7 @@ impl Stream {
         };
 
         if generation_number >= self.values.len() {
-            return Err(UncatchableError::StreamNotContainNeededGeneration {
+            return Err(UncatchableError::StreamDontHaveSuchGeneration {
                 stream: self.clone(),
                 generation,
             }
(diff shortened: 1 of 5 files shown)


air/src/execution_step/errors/uncatchable_errors.rs
```diff
@@ -99,6 +102,13 @@ pub enum UncatchableError {
     /// and not having any CID is considered a non-catching error.
     #[error("{0} for CID {1:?} not found")]
     ValueForCidNotFound(&'static str, String),
+
+    /// Errors occurred while insertion of a value inside stream that doesn't have corresponding generation.
+    #[error(
+        "stream doesn't have generation with number {generation}, supplied to the interpreter data is corrupted,\n\
+             stream is {stream}"
```
monoid, line 109: Shouldn't it be `:?`?

air/src/execution_step/boxed_value/stream.rs
```diff
@@ -327,6 +331,15 @@ impl fmt::Display for Stream {
     }
 }
 
+impl fmt::Display for Generation {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Generation::Nth(generation) => write!(f, "{}", generation),
+            Generation::Last => write!(f, "last"),
```
monoid, line 338: The error below will be formatted as "stream doesn't have generation with number last, ..." that look strange.  Perhaps, Capitalizing the word would be better.

monoid:
utACK

monoid (approved):

PR #493, opened 2023-03-02
chore(ci): Add codecov

nahsi:
covecov check result is not added for some reason. But maybe if merge it will start to work.
https://coveralls.io/github/fluencelabs/aquavm

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,7 @@ jobs:
             RUSTFLAGS="$RUSTFLAGS -Z sanitizer=$san" cargo test --features test_with_native_code --target x86_64-unknown-linux-gnu
           done
 
-      - name: Run codecod
+      - name: Run codecov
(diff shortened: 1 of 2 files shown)


raftedproc (approved):
Tested ACK

.github/workflows/ci.yml
```diff
@@ -31,24 +39,74 @@ jobs:
...
 
+      - name: Setup grcov
+        uses: SierraSoftworks/setup-grcov@v1
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          version: latest
+
+      - name: Run grcov
+        id: coverage
+        uses: actions-rs/grcov@v0.1
```
monoid, line 68: Is it OK that grcov collects data from multiple runs?
It should to get data from `cargo test --features test_with_native_code` only.  I don't know if sanitizer affect coverage report, probably not.

PR #500, opened 2023-03-07
chore(execution-engine) This commit adds a dedicated triplet resolution error for call

raftedproc:


air/src/execution_step/errors/catchable_errors.rs
```diff
@@ -86,6 +86,13 @@ pub enum CatchableError {
     /// This error type is occurred when the length functor applied to a value of non-array type.
     #[error("the length functor could applied only to an array-like value, but it's applied to '{0}'")]
     LengthFunctorAppliedToNotArray(JValue),
+
+    /// Call gets non-string JValue resolving triplet parts.
+    #[error("call cannot resolve non-String triplet variable part `{variable_name}` with value '{actual_value}'")]
+    NonStringJValueInTripletResolution {
```
mikevoronov, line 92: let call it `NonStringValueInTripletResolution` (without `J`)

mikevoronov (approved):
utACK

air/src/execution_step/errors/catchable_errors.rs
raftedproc: Done

PR #501, opened 2023-03-07
feat(interpreter-data)!: New data format for calls

monoid:
BREAKING CHANGE:
1. Call values in the trace have CID references to structures that have call arguments' hash and CID references to values and tetraplets.
2. If call value is unused, it is serialized with `Unused` variant, and CID references are not stored.

Previous data scheme was (Scalar as an example, other cases are similar):

```
Scalar(CID<JValue>) ---<value_store>----> JValue
```

New data scheme is much more sophisticated:

```
Scalar(CID<ServiceResultAggregate>) ---+
                                       |
  +----<service_result_store>----------+
  |
  +-------> ServiceResultAggregate:
               value_cid ------------<value_store>----> JValue
               tetraplet_cid --------<tetraplet_store>----> SecurityTetraplet
               argument_hash: String
```
`Stream` variant is similar, however, `Unused` is different: it has value CID only, but the value is not stored into the `value_store`:

```
Unused(Rc<CID<JValue>>) ---> X
```


crates/air-lib/interpreter-data/src/executed_state.rs
```diff
@@ -53,19 +57,67 @@ pub enum CallResult {
...
+        cid: Rc<CID<ServiceResultAggregate>>,
         generation: u32,
     },
+    /// The call value is not stored.
+    Unused(Rc<CID<ServiceResultAggregate>>),
+}
+
+/// Please note that second argument is a JSON-serialized value, not a plain error message.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct CallServiceFailed(pub i32, pub Rc<String>);
```
mikevoronov, line 100: Let's name fields

crates/air-lib/interpreter-data/src/executed_state.rs
```diff
@@ -39,7 +39,11 @@ pub struct ParResult {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum Sender {
     PeerId(Rc<String>),
-    PeerIdWithCallId { peer_id: Rc<String>, call_id: u32 },
+    PeerIdWithCallId {
+        peer_id: Rc<String>,
+        call_id: u32,
+        argument_hash: Rc<str>,
```
mikevoronov, line 45: Why do we need an argument hash here?

crates/air-lib/interpreter-data/src/executed_state.rs
```diff
@@ -53,19 +57,67 @@ pub enum CallResult {
...
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+/// A proof of service result execution result.
+pub struct ServiceResultAggregate {
+    pub value_cid: Rc<CID<JValue>>,
+    /// Hash of the call arguments.
+    pub argument_hash: Rc<str>,
```
mikevoronov, line 118: Why it's `str`, not byte slice?

crates/air-lib/interpreter-data/src/executed_state/se_de.rs
```diff
@@ -96,19 +100,23 @@ pub mod sender_serializer {
             }
 
             fn visit_str<E: serde::de::Error>(self, raw_sender: &str) -> Result<Self::Value, E> {
```
mikevoronov, line 102: I think we can remove this custom se/de, it was introduced to prevent breaking change in AVM data when `PeerWithId` was introduced.

crates/air-lib/interpreter-data/src/executed_state.rs
monoid: I use CID for the argument hash; also, string serializes better.

air/src/execution_step/execution_context/context.rs
```diff
@@ -146,12 +148,36 @@ pub(crate) struct ExecCtxIngredients {
...
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn insert_value(
+        &mut self,
+        value: Rc<JValue>,
+        tetraplet: RcSecurityTetraplet,
+        argument_hash: Rc<str>,
+    ) -> Result<Rc<CID<ServiceResultAggregate>>, CidCalculationError> {
```
mikevoronov, line 167: Hmm, just noticed that, but looks like `CidCalculationError` isn't really about calculation and more about se/de issues? Maybe it's worth renaming it?

air/src/execution_step/execution_context/context.rs
```diff
@@ -146,12 +148,36 @@ pub(crate) struct ExecCtxIngredients {
...
+        let value_cid = self.value_tracker.record_value(value)?;
+
+        let tetraplet_cid = self.tetraplet_tracker.record_value(tetraplet)?;
```
mikevoronov, line 170:
```suggestion
        let value_cid = self.value_tracker.record_value(value)?;
        let tetraplet_cid = self.tetraplet_tracker.record_value(tetraplet)?;
```

air/src/execution_step/instructions/call/call_result_setter.rs
```diff
@@ -52,11 +58,17 @@ pub(crate) fn populate_context_from_peer_service_result<'i>(
...
-            Ok(CallResult::executed_stream(cid, generation))
+            Ok(CallResult::executed_stream(service_result_agg_cid, generation))
+        }
+        CallOutputValue::None => {
+            let mut dummy_cid_state = ExecutionCidState::new();
+            let service_result_agg_cid = dummy_cid_state
+                .insert_value(executed_result.result, tetraplet, argument_hash)
+                .map_err(UncatchableError::from)?;
+
+            // TODO: Might we add only value's CID for the `Unused`?
```
mikevoronov, line 69: I think yes because tetraplets and hash arguments can't be used in this case; moreover, creating a dummy `ExecutionCidState` takes time and consumes memory for hashmaps.

raftedproc:
UtACK

raftedproc (approved):

crates/air-lib/interpreter-data/src/executed_state.rs
monoid: As discussed: it was intended to save from undefined behavior in variable changing values in `par` branches, but it was suggested to not to do it to reduce data size, moreover, this undefined behavior prevents other peers to verify hashes, and requires other solution.

air/src/execution_step/execution_context/context.rs
monoid: Wouldn't it mean a leaking abstraction then?

mikevoronov: I got this intention, let's leave it as is.

air/src/execution_step/instructions/call/prev_result_handler.rs
```diff
@@ -49,19 +52,37 @@ pub(super) fn handle_prev_state<'i>(
     match met_result.result {
         // this call was failed on one of the previous executions,
         // here it's needed to bubble this special error up
-        CallServiceFailed(ret_code, ref err_msg) => {
+        Failed(ref failed_cid) => {
+            fn deserialize_from_value<T: serde::de::DeserializeOwned>(value: JValue) -> ExecutionResult<T> {
```
mikevoronov, line 56: Why is a separate function needed?

air/src/execution_step/instructions/call/prev_result_handler.rs
```diff
@@ -40,6 +42,7 @@ pub(crate) struct StateDescriptor {
 pub(super) fn handle_prev_state<'i>(
     met_result: MetCallResult,
     tetraplet: &RcSecurityTetraplet,
+    argument_hash: Option<&Rc<str>>,
```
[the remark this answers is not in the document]
monoid: Unfortunately, it doesn't exists for all possible `met_result`.  Fixing it would require substantial refactoring.

air/src/execution_step/instructions/call/prev_result_handler.rs
monoid: It can be removed now.

air/src/execution_step/instructions/call/prev_result_handler.rs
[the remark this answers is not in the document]
mikevoronov: Ok, let's stick with the current variant and refactor it in the future

crates/air-lib/interpreter-data/src/executed_state.rs
```diff
@@ -46,26 +45,72 @@ pub enum Sender {
...
+    ///
+    /// The `JValue` has to be a two element array `[i32, String]`.
+    Failed(Rc<CID<ServiceResultAggregate>>),
 }
 
+/*
+ * The current value structure is:
+ *
+ * ```
+ * Scalar(CID<ServiceResultAggregate>) ---+
```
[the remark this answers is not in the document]
monoid: Done.

PR #503, opened 2023-03-09
chore(execution-engine): Some stream-related LambdaError are unjoinab…

raftedproc:
…le b/c: canon stream replaces normal stream, when canon stream  is used, it is materialized and its size is known

### DIFF
--- a/air/src/execution_step/errors/catchable_errors.rs
+++ b/air/src/execution_step/errors/catchable_errors.rs
@@ -141,15 +141,6 @@ impl Joinable for CatchableError {
                 log_join!("  waiting for an argument with name '{}'", var_name);
                 true
             }
-            LambdaApplierError(LambdaError::StreamNotHaveEnoughValues { stream_size, idx }) => {
-                log_join!("  waiting for an argument with idx '{}' on stream with size '{}'", idx, stream_size);
-                true
-            }
-            LambdaApplierError(LambdaError::EmptyStream) => {
-                log_join!("  waiting on empty stream for path ");
-                true
-            }
-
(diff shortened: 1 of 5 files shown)


mikevoronov (approved):
utACK.

Let's delete these two tests; they no longer test anything.

crates/air-lib/trace-handler/src/data_keeper/trace_slider.rs
```diff
@@ -83,6 +83,7 @@ impl TraceSlider {
     }
 
     pub(crate) fn set_subtrace_len(&mut self, subtrace_len: usize) -> KeeperResult<()> {
+        println!("set_subtrace_len subtrace_len: {}", subtrace_len);
```
mikevoronov, line 86: don't forget to delete it before merging

crates/air-lib/test-utils/src/lib.rs
```diff
@@ -166,5 +166,12 @@ pub fn is_interpreter_succeded(result: &RawAVMOutcome) -> bool {
 }
 
 pub fn check_error(result: &RawAVMOutcome, error: impl ToErrorCode + ToString) -> bool {
+    println!(
```
mikevoronov, line 169: don't forget to delete it before merging

crates/air-lib/trace-handler/src/data_keeper/trace_slider.rs
raftedproc: Those lines accedently slipped into the patch. Thx for your vigilance.

PR #540, opened 2023-03-20
feat(aquavm-air-cli)!: Usability enhancements

monoid:
1. `air run --plain --data` is renamed to `… --current-data` and made optional.
2. if AIR script is read from stdin, a message is printed.  Otherwise use may wait indefinitely if he/she doesn't know that script is read from stdin by default.
3. fix release-please config.

### DIFF
--- a/tools/cli/air/src/trace/run/data/plain.rs
+++ b/tools/cli/air/src/trace/run/data/plain.rs
@@ -46,8 +46,8 @@ pub(crate) struct PlainDataArgs {
 pub(crate) fn load(args: &PlainDataArgs) -> anyhow::Result<ExecutionData<'_>> {
     use super::super::load_data;
 
-    let air_script =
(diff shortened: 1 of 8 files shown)


raftedproc:
utACK

raftedproc (approved):

tools/cli/air/src/trace/run/data/plain.rs
```diff
@@ -84,10 +83,27 @@ fn read_air_script(air_input: Option<&Path>) -> anyhow::Result<String> {
...
             stdin.read_to_string(&mut buffer)?;
             buffer
         }
     };
 
     Ok(air_script)
 }
+
+fn print_air_prompt() {
+    use termcolor::{ColorChoice, ColorSpec, StandardStream, WriteColor as _};
```
mikevoronov, line 100: Can you check does it work on Windows?

mikevoronov (approved):
utACK

PR #569, opened 2023-04-07
chore(trace-handler): sub/-trace len dedicated alias to replace usize [fixes VM-282]

raftedproc:
chore(trace-handler): sub/-trace len dedicated alias to replace usize [fixes VM-282]


crates/air-lib/interpreter-data/src/trace.rs
```diff
@@ -42,6 +46,13 @@ impl ExecutionTrace {
     pub fn push(&mut self, value: ExecutedState) {
         self.0.push(value);
     }
+
+    pub fn trace_states_count(&self) -> TraceLen {
+        self.0
+            .len()
+            .try_into()
+            .expect("trace_states_count() TraceLen failed.")
```
monoid, line 54: One should clearly state that it is an overflow.

crates/air-lib/trace-handler/src/data_keeper/trace_slider.rs
```diff
@@ -102,7 +105,7 @@ impl TraceSlider {
         self.position
     }
 
-    pub(crate) fn subtrace_len(&self) -> usize {
+    pub(crate) fn subtrace_len(&self) -> TraceLen {
         self.subtrace_len - self.seen_elements
```
monoid, line 109: It worth adding a `debug_assert!(self.subtrace_len >= self.seen_elements);`.

crates/air-lib/interpreter-data/src/trace.rs
raftedproc: Done

crates/air-lib/trace-handler/src/data_keeper/trace_slider.rs
raftedproc: Not sure if this would help to catch bugs but added.

air/tests/test_module/negative_tests/uncatchable_trace_related.rs
```diff
@@ -576,13 +576,102 @@ fn fold_pos_overflow() {
...
+            (fold $s i
+                (call "vm_peer_id_2" ("" "") [] a)
+                (next i)
+            )
+        )
+    "#
+    );
+    let mut cid_state = ExecutionCidState::new();
+    let value_pos = 1;
+    let wrong_subtrace_len = u32::MAX;
```
monoid, line 604: Can we write here `TraceLen::MAX`? I don't know if it would work.

air/tests/test_module/negative_tests/uncatchable_trace_related.rs
```diff
@@ -576,13 +576,102 @@ fn fold_pos_overflow() {
...
+        instruction: String::from(String::from("fold $s i")),
+    };
+    assert!(check_error(&result, expected_error));
+}
+
+#[test]
+fn fold_incorrect_subtraces_count() {
+    let vm_peer_id_1 = "vm_peer_id_1";
+    let mut peer_vm_1 = create_avm(unit_call_service(), vm_peer_id_1);
+    let script = format!(
```
monoid, line 645: Do you need `format!` here and at other similar places?

crates/air-lib/interpreter-data/src/trace_pos.rs
```diff
@@ -54,3 +54,11 @@ impl TryFrom<usize> for TracePos {
         PosType::try_from(value).map(TracePos)
     }
 }
+
+impl std::ops::Sub<TracePos> for u32 {
```
monoid, line 58: This piece of code doesn't seem to be needed.

monoid: Moreover, it looks suspicious.  Why should we have `(u32 - TracePos): u32`?

crates/air-lib/trace-handler/src/data_keeper/trace_slider.rs
monoid: It will document the intention.

air/tests/test_module/negative_tests/uncatchable_trace_related.rs
```diff
@@ -576,13 +576,102 @@ fn fold_pos_overflow() {
             subtrace_len: wrong_after_subtrace_len,
         },
     );
-    let resolved_fold = ResolvedFold::new(
-        hashmap![value_pos.into() => fold_lore],
-        (wrong_after_subtrace_len + 1) as usize,
-    );
+    let resolved_fold = ResolvedFold::new(hashmap![value_pos.into() => fold_lore], wrong_after_subtrace_len + 1);
```
monoid, line 579: I think the hash map should be formatted as
```
        hashmap!(
            value_pos.into() => fold_lore,
        ),
```
(Note the `( ... )`).

air/tests/test_module/negative_tests/uncatchable_trace_related.rs
raftedproc: This way I can copy-paste parts of the code. IMHO this doesn't make the test code more complicated to read but easy to change.

air/tests/test_module/negative_tests/uncatchable_trace_related.rs
raftedproc: According with the [maplit docs](https://docs.rs/maplit/1.0.2/maplit/) it doesn't matter which parenthesis to use.

raftedproc: Fixed though.

air/tests/test_module/negative_tests/uncatchable_trace_related.rs
raftedproc: It works.

crates/air-lib/interpreter-data/src/trace_pos.rs
raftedproc: Good catch indeed.

air/tests/test_module/negative_tests/uncatchable_trace_related.rs
monoid: It does matter for style.

raftedproc: From my experience the styling holywars are painful and demoralizing so this should be somehow coded into automated styling enforces.

monoid:
Please, change title to feat, not chore.  Chore commits shouldn't touch production code.

PR #574, opened 2023-04-11
feat(execution-engine): introduce resolver

mikevoronov:
Adds `Resolvable` trait and implements it for all members of `ast` namespace.


air/src/execution_step/resolver/resolvable_impl.rs
```diff
@@ -0,0 +1,146 @@
...
+impl Resolvable for JValue {
+    fn resolve(&self, ctx: &ExecutionCtx<'_>) -> ExecutionResult<(JValue, RcSecurityTetraplets)> {
+        let tetraplet = SecurityTetraplet::literal_tetraplet(ctx.run_parameters.init_peer_id.as_ref());
+        let tetraplet = Rc::new(tetraplet);
+
+        Ok((self.clone(), vec![tetraplet]))
+    }
+}
+
+impl Resolvable for Option<LambdaAST<'_>> {
```
monoid, line 72: I think it should be wrapped with some newtype.

mikevoronov: Agree, but let's do it as a separate PR for consistency because this it should be changed in different places in code base

air/src/execution_step/resolver/resolvable_impl.rs
```diff
@@ -0,0 +1,146 @@
...
+
+        Ok((jvalue, tetraplets))
+    }
+}
+
+impl Resolvable for ast::Scalar<'_> {
+    fn resolve(&self, ctx: &ExecutionCtx<'_>) -> ExecutionResult<(JValue, RcSecurityTetraplets)> {
+        let value = ctx.scalars.get_value(self.name)?.into_jvaluable();
+        let tetraplets = value.as_tetraplets();
+        Ok((value.into_jvalue(), tetraplets))
```
monoid, line 100: Am I right that every time a value is fetched from a variable, it is cloned?
Looks like we can improve it (but that's for another PR, of course).

monoid (approved):
utACK

air/src/execution_step/resolver/resolvable_impl.rs
```diff
@@ -0,0 +1,146 @@
...
+) -> ExecutionResult<(JValue, RcSecurityTetraplets)> {
+    let jvalue = arg.into();
+    let tetraplet = SecurityTetraplet::literal_tetraplet(ctx.run_parameters.init_peer_id.as_ref());
+    let tetraplet = Rc::new(tetraplet);
+
+    Ok((jvalue, vec![tetraplet]))
+}
+
+impl Resolvable for JValue {
+    fn resolve(&self, ctx: &ExecutionCtx<'_>) -> ExecutionResult<(JValue, RcSecurityTetraplets)> {
```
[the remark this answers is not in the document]
mikevoronov: Agree with you, probably we should wrap JValue into smth for that. But for this PR I'll remove it because all work done into the `prepare_const` function.

air/src/execution_step/resolver/resolvable_impl.rs
mikevoronov: Yep, this problem is derived from the previous resolving design, and I wanted to address it in the boxed value project.

PR #578, opened 2023-04-13
fix(execution-engine): Fold-over-scalar values' wrong lambda

monoid:
Previously, iterator values in fold over scalar inherited tetraplet from
the scalar.  But for security guarantees, they should have `.[N]` lambda
added, where N is an element index.


raftedproc (approved):

air/src/execution_step/boxed_value/iterable/lambda_result.rs
```diff
@@ -57,7 +57,9 @@ impl<'ctx> Iterable<'ctx> for IterableLambdaResult {
         }
 
         let jvalue = &self.jvalues[self.cursor];
-        let result = IterableItem::RefRef((jvalue, &self.tetraplet, 0.into()));
+        let mut tetraplet = (*self.tetraplet).clone();
```
mikevoronov, line 60: Should we do the same for every member for which `Iterable` is implemented?

air/tests/test_module/features/tetraplets/security_tetraplets.rs
```diff
@@ -16,6 +16,7 @@
 
 use air::SecurityTetraplet;
 use air_test_utils::prelude::*;
+use pretty_assertions::assert_eq;
```
mikevoronov, line 19: let's shuffle it

mikevoronov (approved):
utACK

air/tests/test_module/features/tetraplets/security_tetraplets.rs
monoid: How can we shuffle it? Both are `other imports from non-std` (not super/crate).

mikevoronov: It's good, I misread it.

PR #579, opened 2023-04-14
chore: rust channel update dedicated to Yuri's Night

raftedproc:


monoid (approved):
utACK

air/src/execution_step/instructions/call/prev_result_handler.rs
```diff
@@ -151,11 +151,11 @@ fn update_state_with_service_result<'i>(
     Ok(())
 }
 
-fn handle_service_error<'i>(
```
mikevoronov, line 154: Was it made by accident? Anyway, I suggest making a separate PR for such changes to keep updates consistent.

monoid: If it was caused by updated clippy warnings, it would break master if not updated.

PR #593, opened 2023-05-04
feat(avm-server,air-interpreter,aquavm-air)!: secret key and particle ID arguments

monoid:
Add `&fluence_keypair::KeyPair` last argument to `AVM::call` and
`AVMRunner::call`.  This value is further forwarded in deconstructed
form to WASM Air interpreter, but is not used there yet.

feat(air-interpreter)!: `invoke` methods have new arguments:
`key_format: u8`, `secret_key_bytes: Vec<u8>` and `particle_id: String`.

feat(aquavm-air)!: `air::execute_air` has two new arguments: `key_format: u8`, `secret_key_bytes: Vec<u8>` and `particle_id: String`.

feat(aquavm-air-cli)!: add `--random-key`/`--ed25519-key` and `--particle-id` options to `run` CLI.


air/src/runner.rs
```diff
@@ -31,6 +31,8 @@ pub fn execute_air(
     data: Vec<u8>,
     params: RunParameters,
     call_results: Vec<u8>,
+    _key_format: u8,
+    _key_bytes: Vec<u8>,
```
folex, line 35: I would expect these fields to be in `RunParameters`, because PeerIds are there.

avm/server/src/runner.rs
```diff
@@ -76,7 +77,13 @@ impl AVMRunner {
         ttl: u32,
         current_peer_id: impl Into<String>,
         call_results: CallResults,
+        keypair: &KeyPair,
     ) -> RunnerResult<RawAVMOutcome> {
+        let key_format = keypair.key_format();
+        let key_bytes: Vec<u8> = keypair
+            .secret()
+            .unwrap_or_else(|_| panic!("failed to serialize key of format {:?}", key_format));
```
folex, line 85: why panic if you can return `Err`? I wouldn't expect `AVM::call` to ever panic

crates/air-lib/test-utils/src/test_runner.rs
```diff
@@ -39,20 +41,40 @@ pub trait AirRunner {
...
 }
 
-#[derive(Debug, Default, Clone)]
+#[derive(Default, Clone)]
 pub struct TestRunParameters {
     pub init_peer_id: String,
     pub timestamp: u64,
     pub ttl: u32,
-    pub override_current_peer_id: Option<String>,
+    pub override_current_peer_id: Option<(String, KeyPair)>,
```
folex, line 59: wow, why use tuple and not add a new field for key pair?

field is called `..._peer_id`, but contains String and KeyPair. That makes me dizzy, code below becomes pretty hard to understand, eg
```
        let key_pair = override_current_peer_id
             .as_ref()
             .map(|pair| &pair.1)
             .unwrap_or(&self.key_secret);
         let override_current_peer_id = override_current_peer_id
             .clone()
             .map(|(peer_id, _key)| peer_id);
```

crates/air-lib/test-utils/src/test_runner.rs
```diff
@@ -77,6 +99,14 @@ impl<R: AirRunner> TestRunner<R> {
         let mut call_results = HashMap::new();
         let mut next_peer_pks = HashSet::new();
 
+        let key_pair = override_current_peer_id
```
folex, line 102: `let key_pair = peer_id.map( pair.1 ).or(key_secret)`

the naming is hard to follow, really. Is it `secret` or `key pair`? or `peer_id`? is `pair` == `keypair` or `tuple`?

avm/server/src/runner.rs
monoid: It panics only if RSA key is provided. OTOH you are right, it is implementation detail.

We might remove RSA key support from `fluence-keypair`, making this method infallible.

crates/air-lib/test-utils/src/test_runner.rs
monoid: Renamed consistently to `keypair`, `keypair_format`, `keypair_data`.

folex: So AVM will have access to both Public and Private key as opposed to just Private key.

If that aligns with the goals, then it's great! :)

monoid: Yes, that's what the goal is, as it puts its key into data.

folex:
It seems that we now pass PublicKey twice: first as `current_peer_id`, then as a part of `KeyPair`.

Btw, what should be passed – Peer Root Key Pair, or Worker Keypair?

If the latter (WK), then it would not match `current_peer_id`, is that OK?

monoid:
> It seems that we will now pass PublicKey twice: first as `current_peer_id`, then as a part of `KeyPair`.

That's true.  It was discussed on a call few days ago.
 
> Btw, what should be passed – Peer Root Key Pair, or Worker Keypair?
> 
> If the latter (WK), then it would not match `current_peer_id`, is that OK?

It has always match `current_peer_id`.  Don't you pass other `current_peer_id` for the worker?

folex:
> That's true. It was discussed on a call few days ago.

What was the outcome? :)

> It has always match `current_peer_id`. Don't you pass other `current_peer_id` for the worker?

So it is OK that signatures will be by Worker Keys? Will we have enough info on chain to associate it with Provider?

monoid:
I believe it should be further discussed.

tools/cli/air/src/trace/run/data/plain.rs
```diff
@@ -60,7 +63,10 @@ pub(crate) fn load(args: &PlainDataArgs) -> anyhow::Result<ExecutionData<'_>> {
 
     let particle = ParticleParameters::new(
         init_peer_id.into(),
-        "".into(),
+        args.particle_id
+            .as_ref()
+            .map(Into::into)
+            .unwrap_or("".into()),
```
mikevoronov, line 69: It looks a bit more efficient
```rust
args.particle_id
            .as_deref()
            .unwrap_or_default()
            .into(),
```

crates/air-lib/interpreter-interface/src/run_parameters.rs
```diff
@@ -39,15 +39,39 @@ pub struct RunParameters {
 
     /// TTL set by init peer id in milliseconds.
     pub ttl: u32,
+
+    /// A key format.
+    ///
+    /// This value is the result of `fluence_keypair::KeyType::into`.
+    pub key_format: u8,
```
mikevoronov, line 46: new line

crates/air-lib/interpreter-interface/src/run_parameters.rs
```diff
@@ -39,15 +39,39 @@ pub struct RunParameters {
...
+    /// A secret key data.
```
mikevoronov, line 47: `data` looks not so clear in this context, let's omit it.

mikevoronov (commented):
We must check that the submitted secret key has the ed25519 format.

crates/air-lib/test-utils/src/test_runner.rs
```diff
@@ -184,14 +226,29 @@ mod tests {
             "func".to_owned() => json!("success"),
         };
 
+        let key_format = fluence_keypair::KeyFormat::Secp256k1;
```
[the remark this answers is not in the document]
monoid: Just borrowed from the `nox` tests.

monoid:
@mikevoronov do you mean we should reject other key types?

folex (approved):

PR #598, opened 2023-05-10
feat(aquavm-air,air-interpreter-signature,air-interpreter-data)!: Peer signatures

monoid:
A peer signs call results and canon results it has produced.

New field `signatures`, a map from peer public key to signature, is added to the interpreter data.

Signatures checking is yet to be done.

### DIFF
--- a/air/src/execution_step/execution_context/context.rs
+++ b/air/src/execution_step/execution_context/context.rs
@@ -75,9 +75,14 @@ pub(crate) struct ExecutionCtx<'i> {
     pub(crate) cid_state: ExecutionCidState,
 
     /// Signatures' store.
+    ///
+    /// It contains peers' signatures for verification.
     pub(crate) signature_store: SignatureStore,
 
-    /// Local signature tracker.
+    /// Local signatures tracker.
+    ///
+    /// It gathers peers' CIDs (call results and canon results) stored in the trace either for signing (current peer's
+    /// CIDs) or sign verification (other peers).
     pub(crate) signature_tracker: SignatureTracker,
 }
 
--- a/air/src/execution_step/execution_context/context.rs
+++ b/air/src/execution_step/execution_context/context.rs
@@ -21,11 +21,14 @@ use super::Scalars;
 use super::Streams;
 
 use air_execution_info_collector::InstructionTracker;
+use air_interpreter_cid::CID;
 use air_interpreter_data::CidInfo;
 use air_interpreter_data::GlobalStreamGens;
 use air_interpreter_data::RestrictedStreamGens;
+use air_interpreter_data::ServiceResultCidAggregate;
 use air_interpreter_interface::*;
 use air_interpreter_signatures::SignatureStore;
+use air_interpreter_signatures::SignatureTracker;
 
 use std::rc::Rc;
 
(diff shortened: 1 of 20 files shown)


air/src/execution_step/execution_context/context.rs
```diff
@@ -69,6 +73,12 @@ pub(crate) struct ExecutionCtx<'i> {
 
     /// CID-to-something trackers.
     pub(crate) cid_state: ExecutionCidState,
+
+    /// Signatures' store.
```
mikevoronov, line 77: let's add something about why is this field needed

air/src/execution_step/execution_context/context.rs
```diff
@@ -69,6 +73,12 @@ pub(crate) struct ExecutionCtx<'i> {
...
+    pub(crate) signature_store: SignatureStore,
+
+    /// Local signature tracker.
```
mikevoronov, line 80: ditto

crates/air-lib/interpreter-signatures/src/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+use std::borrow::Borrow;
+use std::collections::HashMap;
+use std::hash::Hash;
+
+/// An opaque serializable representation of public key.
+///
+/// It can be string or binary, you shouldn't care about it unless you change serialization format.
+#[derive(Debug, Hash, Clone, Eq, PartialEq, PartialOrd, Deserialize, Serialize)]
+#[serde(transparent)]
+pub struct PublicKey(Box<str>);
```
mikevoronov, line 44: let use `Vec<u8>` here

crates/air-lib/interpreter-signatures/src/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+
+impl From<fluence_keypair::PublicKey> for PublicKey {
+    fn from(value: fluence_keypair::PublicKey) -> Self {
+        Self(bs58::encode(&value.to_vec()).into_string().into())
+    }
+}
+
+#[derive(Debug, Hash, Clone, Eq, PartialEq, PartialOrd, Deserialize, Serialize)]
+#[serde(transparent)]
+pub struct Signature(Box<str>);
```
mikevoronov, line 54: ditto

mikevoronov (approved):
utACK

crates/air-lib/interpreter-signatures/src/lib.rs
monoid: They are used as keys in JSON hash maps, and only strings can be used as keys.
This structure is designed to be an opaque type with implemtation can be changed to `Vec<u8>` (or `Box<[u8]>`) when we switch to a binary format.

crates/air-lib/interpreter-signatures/src/lib.rs
monoid: As we use JSON yet, a string is more effective.  The type is designed to be opaque, and can be easily changed to `Vec<u8>` (or `Box<[u8]>`) when we switch to binary format.

crates/air-lib/interpreter-signatures/src/lib.rs
```diff
@@ -0,0 +1,160 @@
...
+    unused_variables,
+    unused_unsafe,
+    unreachable_patterns
+)]
+
+use air_interpreter_cid::CID;
+use fluence_keypair::error::SigningError;
+use fluence_keypair::KeyPair;
+use rand_chacha::rand_core::SeedableRng;
+use serde::{Deserialize, Serialize};
```
raftedproc, line 33: IIRC there should be one import per line.

crates/air-lib/interpreter-signatures/src/lib.rs
mikevoronov: I see, so let's add a comment about that.

PR #599, opened 2023-05-18
fix(aquavm-air): Include unused call results to signature CIDs set

monoid:
While it is not used and can be arbitrary, including this CID to
a signature set prevents this state from being dropped or changed
to request-sent-by state by a malicious peer.


monoid:
IMPORTANT: It will make impossible to statically verify data's signatures.

PR #600, opened 2023-05-19
feat(aquavm-air-cli,performance_metering): memory size in performance report

monoid:
+ aquavm-air-cli outputs final WASM memory size in log.
+ performance_metering collects these memory sizes and reports min and max values for each benchmark.


raftedproc (approved):
utACK

raftedproc (approved):

PR #605, opened 2023-05-31
feat(air-interpreter-signatures): use (de)serialize_with

monoid:
Wrapper types do not hold strings anymore, but use `(de)serialize_with` serde attributes to parse keys and signatures during parse time.

`PublicKey` and `Signature` traits implement `Deref` trait that returns the inner value.


mikevoronov (approved):
utACK

raftedproc (approved):
utACK

PR #608, opened 2023-06-08
chore(execution-engine): additional StreamMap positive tests [fixes VM-295]

raftedproc:
chore(execution-engine): Additional StreamMap positive tests [fixes VM-295]


crates/air-lib/test-utils/src/call_services.rs
```diff
@@ -93,12 +93,14 @@ pub type ArgTetraplets = Vec<Vec<SecurityTetraplet>>;
 
 pub fn tetraplet_host_function(
     closure: impl Fn(CallRequestParams) -> CallServiceResult + 'static,
-) -> (CallServiceClosure, Rc<RefCell<ArgTetraplets>>) {
-    let arg_tetraplets = Rc::new(RefCell::new(ArgTetraplets::new()));
+) -> (CallServiceClosure, Rc<RefCell<Vec<ArgTetraplets>>>) {
```
mikevoronov, line 96: It's better to leave a signature as is because all call services in this file are designed to reflect one single service call. I think it's better to collect tetraplets in the `closure` supplied to this call service.

mikevoronov (approved):
utACK

PR #609, opened 2023-06-16
chore(testing-framework)!: restore WASM test executor

monoid:
Native mode was used before because some package used native runner for its tests.

This PR allows to explicitly select test runner for tests.  Many testing-framework types are now parametrized with a runner type with almost compatible defaults.


crates/testing-framework/src/execution/mod.rs
```diff
@@ -21,22 +21,43 @@ use crate::{
...
     queue: ExecutionQueue,
 }
 
-impl AirScriptExecutor {
+// it is implemented only for the default runner for compatibility reasons
+// Rust fails to deduce type for `AirScriptExecutor::simple()` without
+//   extencive test code changes
+impl AirScriptExecutor<DefaultAirRunner> {
+    /// Simple constructor where everything is generated from the annotated_air_script.
+    pub fn simple(
```
mikevoronov, line 44: I think that `from_annotated` is more readable.

mikevoronov (approved):
utACK

PR #613, opened 2023-06-23
feat!: set nodejs target for air-beautify

shamsartem:
BREAKING CHANGE: this package is now intended for use in nodejs


nahsi (approved):

tools/wasm/air-beautify-wasm/src/lib.rs
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 Fluence Labs Limited
+ * Copyright 2023 Fluence Labs Limited
```
mikevoronov, line 2: In AquaVM/Marine year in copyright, reflect the year when this file was first created (for historical reasons).

mikevoronov (approved):
utACK

tools/wasm/air-beautify-wasm/src/lib.rs
shamsartem: this change is needed to trigger CHANGELOG generation. You can always just use git to learn when file was created. I can revert this and change something else if you think it's important

monoid: The correct value is 2022 :)

mikevoronov: Looks so, ok, let's keep 2022

shamsartem, line 2:
```suggestion
 * Copyright 2022 Fluence Labs Limited
```

PR #618, opened 2023-06-27
feat(air-parser): canon stream syntax

raftedproc:
Maps proposal adds a new way to access canon streams and maps using # sign prefix, e.g. #$canon_stream, #%canon_map.  This patch adds support for # prefix accessing canon structures.

### DIFF
--- a/crates/air-lib/air-parser/src/parser/lexer/call_variable_parser.rs
+++ b/crates/air-lib/air-parser/src/parser/lexer/call_variable_parser.rs
@@ -30,11 +30,12 @@ pub(super) fn try_parse_call_variable(
     CallVariableParser::try_parse(string_to_parse, start_pos)
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone, Copy)]
 enum MetTag {
     None,
     Stream,
     StreamMap,
+    Canon,
     CanonStream,
 }
 
@@ -175,7 +176,10 @@ impl<'input> CallVariableParser<'input> {
     }
 
     fn try_parse_as_variable(&mut self) -> LexerResult<()> {
-        if self.try_parse_as_stream_start()? || self.try_parse_as_json_path_start()? {
+        if self.try_parse_as_canon()?
+            || self.try_parse_as_stream()?
+            || self.try_parse_as_json_path_start()?
+        {
             return Ok(());
         } else if self.is_json_path_started() {
             self.try_parse_as_json_path()?;
@@ -186,15 +190,31 @@ impl<'input> CallVariableParser<'input> {
         Ok(())
     }
 
-    fn try_parse_as_stream_start(&mut self) -> LexerResult<bool> {
-        let stream_tag = MetTag::from_tag(self.current_char());
-        if self.current_offset() == 0 && stream_tag.is_tag() {
+    fn try_parse_as_stream(&mut self) -> LexerResult<bool> {
+        let tag = MetTag::from_tag(self.current_char());
+        if self.current_offset() == 0 && tag.is_tag() {
             if self.string_to_parse.len() == 1 {
                 let error_pos = self.pos_in_string_to_parse();
-                return Err(LexerError::empty_stream_name(error_pos..error_pos));
+                return Err(LexerError::empty_tagged_name(error_pos..error_pos));
             }
 
-            self.state.met_tag = stream_tag;
+            self.state.met_tag = tag;
+            return Ok(true);
+        }
+
+        Ok(false)
+    }
+
+    fn try_parse_as_canon(&mut self) -> LexerResult<bool> {
+        let tag = self.state.met_tag.deduce_tag(self.current_char());
+
+        if self.current_offset() == 1 && tag.is_canon_stream() {
+            if self.string_to_parse.len() == 2 && tag.is_tag() {
+                let error_pos = self.pos_in_string_to_parse();
+                return Err(LexerError::empty_canon_name(error_pos..error_pos));
+            }
+
+            self.state.met_tag = tag;
             return Ok(true);
         }
 
@@ -238,6 +258,9 @@ impl<'input> CallVariableParser<'input> {
                 return Err(LexerError::leading_dot(
                     self.start_pos..self.pos_in_string_to_parse(),
                 ));
+            } else if self.state.met_tag.is_tag() && self.current_offset() <= 2 {
+                let prev_pos = self.pos_in_string_to_parse() - 1;
+                return Err(LexerError::empty_canon_name(prev_pos..prev_pos));
             }
             self.state.first_dot_met_pos = Some(self.current_offset());
             return Ok(true);
@@ -288,7 +311,7 @@ impl<'input> CallVariableParser<'input> {
                 name,
                 position: self.start_pos,
             },
-            MetTag::CanonStream => Token::CanonStream {
+            MetTag::CanonStream | MetTag::Canon => Token::CanonStream {
                 name,
                 position: self.start_pos,
             },
@@ -311,7 +334,7 @@ impl<'input> CallVariableParser<'input> {
                 lambda,
                 position: self.start_pos,
             },
-            MetTag::CanonStream => Token::CanonStreamWithLambda {
+            MetTag::CanonStream | MetTag::Canon => Token::CanonStreamWithLambda {
                 name,
                 lambda,
                 position: self.start_pos,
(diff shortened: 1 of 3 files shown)


crates/air-lib/air-parser/src/parser/lexer/call_variable_parser.rs
```diff
@@ -186,15 +197,30 @@ impl<'input> CallVariableParser<'input> {
...
 
-    fn try_parse_as_stream_start(&mut self) -> LexerResult<bool> {
-        let stream_tag = MetTag::from_tag(self.current_char());
-        if self.current_offset() == 0 && stream_tag.is_tag() {
+    fn try_parse_as_tagged_token(&mut self) -> LexerResult<bool> {
+        let tag = MetTag::from_tag(self.current_char());
+        if self.current_offset() == 0 && tag.is_tag() {
             if self.string_to_parse.len() == 1 {
                 let error_pos = self.pos_in_string_to_parse();
                 return Err(LexerError::empty_stream_name(error_pos..error_pos));
```
raftedproc, line 205: I will also change the error name if you @mikevoronov find the change suitable.

mikevoronov: yep, it seems obsolete

monoid:
What is the intended syntax change?  Please, fill the description.

crates/air-lib/air-parser/src/parser/lexer/call_variable_parser.rs
```diff
@@ -387,12 +413,27 @@ impl MetTag {
...
         match tag {
             '$' => Self::Stream,
-            '#' => Self::CanonStream,
+            '#' => Self::Canon,
             '%' => Self::StreamMap,
             _ => Self::None,
         }
     }
 
+    fn update_type(&self, tag: char) -> Self {
```
monoid, line 422: IMHO should be `update_canon_type`.

crates/air-lib/air-parser/src/parser/lexer/call_variable_parser.rs
```diff
@@ -387,12 +413,27 @@ impl MetTag {
...
+        match tag {
+            '$' => Self::CanonStream,
+            _ => self.to_owned(),
+        }
+    }
+
+    fn is_canon(&self) -> bool {
+        matches!(self, Self::Canon)
```
monoid, line 430: If you derive `PartialEq` on `MetTag`, you may simply write `*self == Self::Canon`.

It worth deriving `Eq` too, as it OK for this type.

crates/air-lib/air-parser/src/parser/lexer/call_variable_parser.rs
```diff
@@ -186,15 +190,34 @@ impl<'input> CallVariableParser<'input> {
         Ok(())
     }
 
-    fn try_parse_as_stream_start(&mut self) -> LexerResult<bool> {
-        let stream_tag = MetTag::from_tag(self.current_char());
-        if self.current_offset() == 0 && stream_tag.is_tag() {
+    fn try_parse_as_tagged_token(&mut self) -> LexerResult<bool> {
```
monoid, line 193: It worth commenting what is "tagged token".

crates/air-lib/air-parser/src/parser/lexer/call_variable_parser.rs
raftedproc: What is a benefit comparing with the current code?

monoid: Simple equality seems to be more straightforward.

crates/air-lib/air-parser/src/parser/lexer/call_variable_parser.rs
monoid: But most likely it will be context-dependent, won't it?

raftedproc: Doesn't look like it. Match + if will do the trick.

crates/air-lib/air-parser/src/parser/lexer/call_variable_parser.rs
raftedproc: I tend to stick to whatever is used now and see no real benefit changing this to equality b/c it is internally the same.

crates/air-lib/air-parser/src/parser/lexer/call_variable_parser.rs
monoid: Then you will need to update all the call site, as you call it now under `if`.  Perhaps, it worth moving that check inside the function right now.

raftedproc: This makes sense.

monoid: Also, in this hypothetical future it is not so hard to rename the function back if it becomes more generic, is it?

raftedproc: I prefer to stick with a generic name now to potentially save some "cycles" in future. I mean it looks like a matter of opinions.

raftedproc: I will rename the function into maybe_change_type and move if into it.

monoid: You should save reader's "cycles", not yours.

crates/air-lib/air-parser/src/parser/lexer/call_variable_parser.rs
monoid: Can you clarify, please?

crates/air-lib/air-parser/src/parser/lexer/call_variable_parser.rs
raftedproc: The future programmer will be the reader/editor whose cycles I am saving. 
maybe_change_type will return a value thus a reader will see the asignment.

crates/air-lib/air-parser/src/parser/lexer/call_variable_parser.rs
raftedproc: Sure. I am going to add comments on this matter. Not sure yet what is a suitable place for the comment.

crates/air-lib/air-parser/src/parser/lexer/call_variable_parser.rs
monoid: IMHO, a method can update state and return something as well.

mikevoronov (changes requested):

crates/air-lib/air-parser/src/parser/lexer/call_variable_parser.rs
```diff
@@ -186,15 +190,34 @@ impl<'input> CallVariableParser<'input> {
...
+    fn try_parse_as_canon(&mut self) -> LexerResult<bool> {
+        let tag = if self.state.met_tag.is_canon() {
+            self.state.met_tag.update_type(self.current_char())
+        } else {
+            self.state.met_tag
+        };
+        if self.current_offset() == 1 && tag.is_canon_stream() {
+            if self.string_to_parse.len() == 2 && tag.is_tag() {
+                let error_pos = self.pos_in_string_to_parse();
+                return Err(LexerError::empty_canon_name(error_pos..error_pos));
```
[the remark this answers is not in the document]
raftedproc: I am going to rename try_parse_as_tagged_token to try_parse_as_stream as was discussed.

crates/air-lib/air-parser/src/parser/lexer/call_variable_parser.rs
monoid: Code style is not about internally at all.

raftedproc: IMHO it is unwise to use Eq here given that is_tag() uses !matches and I am not changing is_tag b/c it is outside the scope of this patch. You suggest to break the style from my point of view.

crates/air-lib/air-parser/src/parser/lexer/call_variable_parser.rs
raftedproc: Added the comments near MetTag implementation.

crates/air-lib/air-parser/src/parser/lexer/call_variable_parser.rs
monoid: Using a PR for small style/idiomatic changes is OK, otherwise they will never be changed.

`matches!` should be used for types that cannot be compared for some reason ("I haven't defined PartialEq just because" is not the reason).  This enum is essentially simple C-style enum.

mikevoronov (approved):
utACK

PR #621, opened 2023-06-30
feat(execution-engine)!: refactor streams [fixes VM-255]

### LINKED ISSUE #642: bug: restricted stream generations are interfered in fold over scalar
Restricted stream generations are accumulated in dedicated hashmaps inside data, this hashmap contains an array of generations for every `new` instruction in code. It relies on the fact that every time order of these values is the same but It causes problems with their ordering in fold over scalars if generations are stored inside current data.

In the following example execution flow could be as follows:
```
(1) peer_id_1 -> (2) peer_id_3 -> (3) peer_id_1 -> (4) peer_id_1
                               ->     peer_id_2
```
Here in the (3) step `peer_id_1` and `peer_id_2` will have the following stream generations in data:
```
<new position> <-> [2]
```

And then in the (4) step, AquaVM tries to execute the `fold` body with generations `[2]` and on the first cycle, both two values from data will be used. Then in the next step, `new` will obtain 0s from previous and current data, it'll cause a crash on adding values by `ap` because it'll try to add it into 1'st generation which the stream doesn't have.

``` wasm
(seq
    (seq
        (seq
             (ap "peer_id_1" $peers)
             (ap "peer_id_2" $peers))
        (canon "peer_id_3" $peers #peers))
    (seq
        (fold #peers peer
             (par
                 (seq
                     (call peer ("" "") [])
                     (new $stream
                         (seq
                              (call peer ("" "") [] $stream)
                              (seq
                                  (call peer ("" "") [] $stream)
                                  (ap 1 $stream)))))
                (next peer)))
        (call "peer_id_1" ("" "") [])))
```

### LINKED ISSUE #644: bug: restricted stream generations are interfered in fold over stream
The idea of this bug is almost the same as in #642, but stream bodies' permutations cause it.

Minimal reporducibale script
```wasm
(seq
    (seq
        (par
             (call %init_peer_id% ("" "") [%init_peer_id%] $peers)
             (call "{peer_id_2}" ("" "") ["{peer_id_2}"] $peers))
        (canon "{peer_id_3}" $peers #peers))
    (seq
        (fold $peers peer
             (seq
                 (seq
                     (call peer ("" "") [""])
                     (new $stream
                         (seq
                              (call peer ("" "") [""] $stream)
                              (seq
                                  (call peer ("" "") [""] $stream)
                                  (ap 1 $stream)))))
                (next peer)))
        (call %init_peer_id% ("" "") [""])))
```

mikevoronov:
Refactored stream and stream generation a lot, it introduces the following changes:
 - no generation in data anymore, AquaVM relies on generation inside data to stay valid and places value accordingly to it
 - stream is internally divided into previous, current, and new values, before, it was one array for all of them
 - recursive streams cursors are refactored and rely on new generation values instead
 - the `Generation` enum was refactored and now contains the source of the generation

Closes #642.
Closes #644.


air/src/execution_step/boxed_value/stream/values_matrix.rs
```diff
@@ -0,0 +1,168 @@
...
+impl<T> ValuesMatrix<T> {
+    pub fn new() -> Self {
+        Self { values: TiVec::new() }
+    }
+
+    pub fn remove_empty_generations(&mut self) {
+        self.values.retain(|generation| !generation.is_empty())
+    }
+
+    pub fn len(&self) -> GenerationIdx {
```
raftedproc, line 44: The name looks misguiding. You ask for len() and return GenerationIdx.

air/src/execution_step/boxed_value/stream/values_matrix.rs
```diff
@@ -0,0 +1,168 @@
...
+            .filter(|generation| !generation.is_empty())
+            .skip(skip.into())
+            .map(|generation| generation.as_ref())
+    }
+}
+
+impl<T: Clone> ValuesMatrix<T> {
+    pub fn add_value_to_generation(&mut self, value: T, generation_idx: GenerationIdx) {
+        if generation_idx >= self.values.len() {
+            let new_size = generation_idx.checked_add(1.into()).unwrap();
```
raftedproc, line 64: It is worth to at least use expect here given that checked_add is supposed to protect against GenerationIdx wraparound.

air/src/execution_step/boxed_value/stream/values_matrix.rs
```diff
@@ -0,0 +1,168 @@
...
+            self.values.resize(new_size.into(), Vec::new());
+        }
+
+        self.values[generation_idx].push(value);
```
raftedproc, line 68: There is an implicit assumption here that generation_idx can be less, equal or bigger by 1 comparing with self.values.len() that either needs to be documented or asserted IMHO.
Consider the example when generation_idx = self.values.len() + 2.

air/src/execution_step/boxed_value/stream/stream_definition.rs
```diff
@@ -0,0 +1,553 @@
...
+        let generation_id = usize::try_from(generation_id).unwrap();
+        let generation_idx = GenerationIdx::from(generation_id);
+        Self::Previous(generation_idx)
+    }
+
+    #[cfg(test)]
+    pub fn current(generation_id: u32) -> Self {
+        use std::convert::TryFrom;
+
+        let generation_id = usize::try_from(generation_id).unwrap();
```
raftedproc, line 161: ditto

crates/air-lib/test-utils/src/test_runner.rs
```diff
@@ -106,6 +106,8 @@ impl<R: AirRunner> TestRunner<R> {
                 )
                 .map_err(|e| e.to_string())?;
 
+            // super::print_trace(&outcome, "");
```
raftedproc, line 109: This seems like a leftover.

air/tests/test_module/features/streams/compactification.rs
```diff
@@ -0,0 +1,121 @@
...
+}
+
+#[test]
+fn global_stream_maps_are_compactified() {
+    let peer_id = "peer_id";
+    let service_result = "service_result";
+    let script = format!(
+        r#"
+        (seq
+            (ap 1 $stream)
```
raftedproc, line 62: The test name implies here must be a %stream_map.

air/src/execution_step/boxed_value/stream/values_matrix.rs
mikevoronov: Do you have any ideas in mind?

air/src/execution_step/boxed_value/stream/values_matrix.rs
mikevoronov: It's a really short function that ensures that a `values` is long enough for supplied `generation_idx`,

air/src/execution_step/boxed_value/stream/values_matrix.rs
mikevoronov: I added TODO to replace this unwrap with error

air/src/execution_step/boxed_value/stream/stream_definition.rs
```diff
@@ -0,0 +1,553 @@
...
+    fn update_generations(
+        values: impl Iterator<Item = &'value [T]>,
+        start_idx: GenerationIdx,
+        trace_ctx: &mut TraceHandler,
+    ) -> ExecutionResult<()> {
+        use crate::execution_step::errors::UncatchableError;
+        use crate::execution_step::ExecutionError;
+
+        for (position, values) in values.enumerate() {
+            let generation = start_idx.checked_add(position.into()).unwrap();
```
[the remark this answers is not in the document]
mikevoronov: the same with TODO

air/src/execution_step/boxed_value/stream/stream_definition.rs
```diff
@@ -0,0 +1,553 @@
...
+    Current(GenerationIdx),
+    New,
+}
+
+impl Generation {
+    #[cfg(test)]
+    pub fn previous(generation_id: u32) -> Self {
+        use std::convert::TryFrom;
+
+        let generation_id = usize::try_from(generation_id).unwrap();
```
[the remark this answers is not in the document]
mikevoronov: I'm not sure because it's worth nothing to change the argument type to `u64`

air/src/execution_step/boxed_value/stream/stream_definition.rs
mikevoronov: I'm not sure because it's worth nothing to change the argument type to `u64`.

air/tests/test_module/features/streams/compactification.rs
mikevoronov: Thx, missed it

air/src/execution_step/boxed_value/stream/values_matrix.rs
mikevoronov: what about `generations_count`?

raftedproc (approved):
utACK

air/src/execution_step/boxed_value/stream/values_matrix.rs
raftedproc: I would call it last_generation but generation_count also works.

PR #622, opened 2023-06-30
feat(execution-engine)!: add error-code invariant check for match/mismatch

raftedproc:
feat(execution-engine)!: add error-code invariant check for match/mismatch

### DIFF
--- a/air/src/execution_step/errors/catchable_errors.rs
+++ b/air/src/execution_step/errors/catchable_errors.rs
@@ -39,6 +39,14 @@ pub enum CatchableError {
     #[error("Local service error, ret_code is {0}, error message is '{1}'")]
     LocalServiceError(i32, Rc<String>),
 
+    /// This error type is produced by a match to notify xor that compared values aren't equal.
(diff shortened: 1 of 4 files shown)


air/src/execution_step/errors/catchable_errors.rs
```diff
@@ -35,6 +35,14 @@ use std::rc::Rc;
...
 #[strum_discriminants(derive(EnumIter))]
 pub enum CatchableError {
+    /// This error type is produced by a match to notify xor that compared values aren't equal.
+    #[error("match is used without corresponding xor")]
+    MatchValuesNotEqual,
+
+    /// This error type is produced by a mismatch to notify xor that compared values aren't equal.
+    #[error("mismatch is used without corresponding xor")]
+    MismatchValuesEqual,
+
```
mikevoronov, line 45: I just thought with a fresh view that it's better to place them after `LocalServiceError`, because the latter is more important error type.

mikevoronov (approved):
utACK

PR #627, opened 2023-07-03
chore: remove fstrings dep

mikevoronov:
Remove the `fstrings` dep because it prevent compilation for RISC0.


raftedproc (approved):

monoid:
There are multiple `f!` usages it tests.  As this package is obsolete, it worth removing it completely.

monoid (approved):
utACK

PR #631, opened 2023-07-12
feat(aquavm-air)!: `ap` join behavior

### LINKED ISSUE #632: bug: ... state ... incompatible with expected `ap`
Certain spell fails from time to time with a error
```
"on instruction 'ap gte $res-0' trace handler encountered an error: state from previous `Call(RequestSentBy(PeerIdWithCallId { peer_id: \"xxxxxx\", call_id: 7 }))` is incompatible with expected ap"
```

Minimal example:

```
(xor
 (seq
  (par
   (call %init_peer_id% ("" "") [0])
   (call %init_peer_id% ("" "") [1] x)
  )
  (ap x $y)
 )
 (call %init_peer_id% ("" "") [2])
)
```

Call execution sequence:

+ in the first invocation, calls 0 and 1 are requested.
+ on the second invocation, call 0 result is provided; it leads to immediate execution of the ap  that fails because x is in not defined, leaving nothing in the trace, but the request of the call 2 is left in the trace.
+ on the third invocation, the result 1 is provided.  When execution reaches the `ap`, it founds the call request of the call 2 issued previously.

monoid:
Previously, `ap` of an undefined (first) variable was a catchable error.
Now `ap` has join behavior for its first argument.

Closes #632


air/src/execution_step/instructions/ap.rs
```diff
@@ -40,10 +42,12 @@ impl<'i> super::ExecutableInstruction<'i> for Ap<'i> {
     fn execute(&self, exec_ctx: &mut ExecutionCtx<'i>, trace_ctx: &mut TraceHandler) -> ExecutionResult<()> {
         log_instruction!(call, exec_ctx, trace_ctx);
         let should_touch_trace = should_touch_trace(self);
-        // this applying should be at the very beginning of this function,
-        // because it's necessary to check argument lambda, for more details see
-        // https://github.com/fluencelabs/aquavm/issues/216
```
mikevoronov, line 45: let's leave the comment

air/tests/test_module/issues/mod.rs
```diff
@@ -31,6 +31,7 @@ mod issue_300;
 mod issue_302;
 mod issue_304;
 mod issue_306;
+mod issue_310;
```
mikevoronov, line 34: let's file an issue and name it accordingly

mikevoronov (approved):
utACK

air/tests/test_module/issues/mod.rs
monoid: The `ap` join behavior issue is VM-316, but the VM-310 issue demonstrates the original problem that folex had.

mikevoronov: This list is based on github issues

mikevoronov: Ok, let me put it in the following way:
 - we have linear issues
 - we still have github issues
 
From the code perspective, list in 'air/tests/test_module/issues/mod.rs' is based on the github issues because they are public. And for this PR I propose to file a new issue in github and use its number here.

monoid: Sorry, I see now.

PR #633, opened 2023-07-14
feat(air-test-utils): `print_trace` prints values

monoid:
As traces contain CIDs for calls and canons, one cannot easily see values produced.  Print them for convenience.


mikevoronov (commented):
utACK

mikevoronov (approved):

PR #635, opened 2023-07-19
chore(execution-engine): move get_unique_map_keys_stream into StreamMap

mikevoronov:
This PR moves the creation of a stream with unique keys from `Stream` to `StreamMap` to improve code isolation. Before this PR `Stream` has known about `StreamMapKey` which wasn't compatible with the dependencies flow.

It is needed in order to support #621.

### DIFF
--- a/air/src/execution_step/boxed_value/stream_map.rs
+++ b/air/src/execution_step/boxed_value/stream_map.rs
@@ -87,6 +87,7 @@ impl StreamMap {
 
         let mut distinct_keys = HashSet::new();
 
+        // unwrap is safe because slice_iter always returns Some iff generation slice is valid
         let new_values = self
             .stream
             .slice_iter(Generation::Nth(0.into()), Generation::Last)
--- a/air/src/execution_step/boxed_value/stream_map.rs
+++ b/air/src/execution_step/boxed_value/stream_map.rs
@@ -86,9 +86,9 @@ impl StreamMap {
     pub(crate) fn create_unique_keys_stream(&self) -> Stream {
         use std::collections::HashSet;
 
-        let mut meet_keys = HashSet::new();
+        let mut met_keys = HashSet::new();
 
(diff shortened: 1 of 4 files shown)


air/src/execution_step/boxed_value/stream_map.rs
```diff
@@ -83,7 +83,30 @@ impl StreamMap {
...
+                        StreamMapKey::from_kvpair(v)
+                            .map(|key| distinct_keys.insert(key))
+                            .unwrap_or(false)
+                    })
+                    .cloned()
+                    .collect::<Vec<_>>()
+            })
+            .collect::<Vec<_>>();
+
+        let stream = Stream::new(new_values, self.stream.previous_gens_count());
```
monoid, line 108: It seems that explicit types are not needed in both `.collect` above, as the type is deduced from `Stream::new` type, buy may be kept for documentation proposes.

mikevoronov: Yep, it's mostly for documentation purposes.

raftedproc (approved):

air/src/execution_step/boxed_value/stream_map.rs
```diff
@@ -82,8 +81,32 @@ impl StreamMap {
...
     }
 
-    pub(crate) fn get_unique_map_keys_stream(&mut self) -> Cow<'_, Stream> {
-        self.stream.get_unique_map_keys_stream()
+    // TODO: change the implementation to mutate the underlying stream
+    // instead of creating a new one
+    pub(crate) fn create_unique_keys_stream(&self) -> Stream {
+        use std::collections::HashSet;
+
+        let mut meet_keys = HashSet::new();
```
monoid, line 89: I think they are `met_keys`.

monoid (approved):
utACK

PR #636, opened 2023-07-19
chore(execution-engine): refactor unseen canon stream creation

mikevoronov:
- refactor canon instruction implementation
- `get_stream_or_default` creates a new `CanonStream` instead of returning a stream 
- `record_value` renamed to `track_value`

In order to support #621.

### DIFF
--- a/air/src/execution_step/boxed_value/stream_map.rs
+++ b/air/src/execution_step/boxed_value/stream_map.rs
@@ -256,12 +256,12 @@ mod test {
             );
         }
 
-        let mut unique_keys_iter = stream_map.iter_unique_key();
+        let mut iter = stream_map.iter_unique_key();
 
-        assert_eq!(&json!(0), unique_keys_iter.next().unwrap().get_result().get("value").unwrap());
-        assert_eq!(&json!(1), unique_keys_iter.next().unwrap().get_result().get("value").unwrap());
-        assert_eq!(&json!(2), unique_keys_iter.next().unwrap().get_result().get("value").unwrap());
-        assert_eq!(unique_keys_iter.next(), None);
+        assert_eq!(&json!(0), iter.next().unwrap().get_result().get("value").unwrap());
+        assert_eq!(&json!(1), iter.next().unwrap().get_result().get("value").unwrap());
+        assert_eq!(&json!(2), iter.next().unwrap().get_result().get("value").unwrap());
+        assert_eq!(iter.next(), None);
     }
 
     #[test]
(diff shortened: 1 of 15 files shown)


air/src/execution_step/instructions/canon.rs
```diff
@@ -14,189 +14,80 @@
...
-        };
-
         match canon_result {
             MergerCanonResult::CanonResult(canon_result_cid) => {
                 handle_seen_canon(epilog, canon_result_cid, exec_ctx, trace_ctx)
             }
             MergerCanonResult::Empty => {
-                let get_stream_or_default = get_stream_or_default_closure(self.stream.name, self.stream.position);
-                handle_unseen_canon(epilog, &get_stream_or_default, &self.peer_id, exec_ctx, trace_ctx)
+                let create_canon_stream = create_canon_stream_closure(self.stream.name, self.stream.position);
```
raftedproc, line 52: Technically create_canon_stream is the closure so I couldn't get from the naming what is what.
```suggestion
                let canon_stream_producer = create_canon_stream_producer(self.stream.name, self.stream.position);
```

air/src/execution_step/instructions/canon_utils/mod.rs
```diff
@@ -0,0 +1,107 @@
+/*
+ * Copyright 2022 Fluence Labs Limited
```
raftedproc, line 2: 2023

air/src/execution_step/instructions/canon_utils/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+    let peer_id = resolve_peer_id_to_string(peer_id, exec_ctx)?;
+
+    if exec_ctx.run_parameters.current_peer_id.as_str() != peer_id {
+        exec_ctx.make_subgraph_incomplete();
+        exec_ctx.next_peer_pks.push(peer_id);
+        //this branch is executed only when
+        //  this canon instruction executes for the first time
+        //  a peer is different from one set in peer_id of a this canon instruction
+        //
+        // the former means that there wasn't canon associated state in data, the latter that it
```
raftedproc, line 72: I might be wrong, but `there was no canon associated state in the data` looks more appropriate here.
I assume the second half should have a verb, e.g. `the latter means that it`

raftedproc (approved):
utACK. There are minor suggestions though.

air/src/execution_step/boxed_value/stream_map.rs
```diff
@@ -81,32 +81,18 @@ impl StreamMap {
...
-                    })
-                    .cloned()
-                    .collect::<Vec<_>>()
-            })
-            .collect::<Vec<_>>();
-
-        Stream::new(new_values, self.stream.previous_gens_count())
+        // it's always possible to go through all values
+        self.stream.iter(Generation::Last).unwrap().filter(move |value| {
+            StreamMapKey::from_kvpair(value)
```
monoid, line 92: `...::from_kvpair` can be optimized, returning `Cow::Borrowed`.

monoid (approved):
utACK

air/src/execution_step/boxed_value/stream_map.rs
mikevoronov: good catch, but lets do it in a separate PR

monoid: No problem.

PR #607, opened 2023-06-07
feat(aquavm-air)!: signature checking

monoid:
- Attacks tests sketch
- Validate CID data corruption
- Some more attacks that are not detected yet
- Trace `CidInfo::verify`
- DRAFT Cross-store data validation
- DRAFT Verifying call parameters on execution
- WIP: InterpreterData verification
- DRAFT: PublicKey is fluence_keypair::PublicKey newtype
- WIP: Use `Verifier` in `air::prepare`
- Track only current peer's CIDs
- Refactor signature trackers to module
- Refactor signature stores
- Refactor AIR stages
- feat(aquavm-air): add `check_signatures` flag
- feat(aquavm-air): add `sign_cids` feature flag
- Fix canon tests
- Correct check of peer's presence
- Refactor interpreter data verification
- Fix tests

PR #637, opened 2023-07-21
chore: run llvm-cov on default features only

monoid:
In the PR #607 tests are not compatible with some feautre flags yet.
Disable `llvm-cov`'s `--all-feautres` until all the tests are fixed.
For now, run `llvm-cov` only with default features.


mikevoronov (approved):
utACK

PR #662, opened 2023-08-07
chore(trace-handler,air-parser): replace matches! with equality comparison where applicable

raftedproc:
Per a discussion raised some time ago this patch replaces matches! with equality comparison where applicable.


air/tests/test_module/features/data_merging/data_merge.rs
```diff
@@ -190,13 +190,13 @@ fn stream_merge() {
     let mut vm2 = create_avm(neighborhood_call_service, "B");
```
mikevoronov, line 190: Let's no change this file in order to make commit history better understandable

mikevoronov (approved):
utACK

monoid (approved):
utACK

PR #675, opened 2023-08-15
chore(testing-framework): clearer triplet error message

monoid:
ATM service ID has to be a static string; make it more explicit in the error message.

### DIFF
--- a/crates/testing-framework/src/transform/parser.rs
+++ b/crates/testing-framework/src/transform/parser.rs
@@ -239,7 +239,7 @@ fn parse_sexp_call_triplet(inp: Input<'_>) -> IResult<Input<'_>, Box<Triplet>, P
             delimited(
                 delim_ws(tag("(")),
                 separated_pair(
-                    context("triplet service name", parse_sexp_string),
+                    context("triplet service name has to be a string", parse_sexp_string),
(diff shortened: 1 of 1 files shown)


crates/testing-framework/src/transform/parser.rs
```diff
@@ -239,7 +239,7 @@ fn parse_sexp_call_triplet(inp: Input<'_>) -> IResult<Input<'_>, Box<Triplet>, P
             delimited(
                 delim_ws(tag("(")),
                 separated_pair(
-                    context("triplet service name", parse_sexp_string),
+                    context("triplet service name has to be a string", parse_sexp_string),
```
raftedproc, line 242: Do you mean a literal string or scalar is also fine as a module name?

raftedproc: If literal is the only option I would change it to `has to be a literal string`.

raftedproc (approved):
utACK

crates/testing-framework/src/transform/parser.rs
monoid: It is only literal string that can be used as a service ID, you are right.

PR #676, opened 2023-08-15
chore: `PartialOrd` uses `Ord`'s implementation

monoid:
It was vice versa.  The problem was found by clippy.


raftedproc:
I am curious what are the reasons the checker leverages suggesting the order(PartialOrd->Ord to Ord->PartialOrd) change?
IMHO it is a matter of preference which comparison trait goes first.

raftedproc (commented):
utACK

monoid:
@raftedproc one `.unwrap()` less.

raftedproc (approved):

PR #679, opened 2023-08-18
fix(aquavm-air-cli): NEAR mode fixes

monoid:
1. Correct default contract path.
2. More informative error messages.
3. Deserialize contract result correctly.


mikevoronov (approved):
utACK

PR #681, opened 2023-08-23
feat(air-cli)!: introduce risc-0 extension

mikevoronov:


tools/cli/air/src/trace/run.rs
```diff
@@ -215,42 +229,38 @@ pub(crate) fn run(args: Args) -> anyhow::Result<()> {
...
-    let mode = mode.unwrap_or(Mode::Native);
-    match mode {
-        Mode::Native => {
-            self::native::create_native_avm_runner().context("Failed to instantiate a native AVM")
-        }
-        #[cfg(feature = "near")]
-        Mode::Near => self::near::create_near_runner(_air_contract_wasm_path)
-            .context("Failed to instantiate NEAR AVM"),
-    }
+fn default_mode() -> Mode {
```
monoid, line 257: If it was a constant, it would be more straightforward.

tools/cli/air/src/trace/run/risc0.rs
```diff
@@ -0,0 +1,117 @@
...
+    let verification_timer = Instant::now();
+    receipt.verify(ZK_AQUAVM_ID)?;
+    let verification_duration = verification_timer.elapsed();
+
+    eprintln!(
+        "verification successfully finished in {:?}",
+        verification_duration
+    );
+
+    from_slice(&receipt.journal).map_err(Into::into)
```
monoid, line 116: AquaVM's `execute_air` returns `InterpreterOutcome`, and decoding it to `RawAVMOutcome` is not correct:
https://github.com/fluencelabs/aquavm/commit/6978603c76a2a2576fc4196369d944f25da7a559#diff-78db3e874ee9b231b30c4c639cc777d57c90c40e3ad9d61365cb05b6069003c1R123-R128

tools/cli/air/src/trace/run.rs
mikevoronov: I don't have any special opinion here, let it be a constant

air/Cargo.toml
```diff
@@ -64,10 +64,8 @@ pretty_assertions = "0.6.1"
 serde_json = "1.0.61"
 
 [features]
-default = ["rand"]
 check_signatures = []
 gen_signatures = []
-rand = ["air-interpreter-data/rand", "air-interpreter-signatures/rand"]
```
monoid: I believe it will break `air-near-contract`.

mikevoronov: I added this line in some commits in this PR and then removed b/c it's doesn't influence RISC-0, though I think it's still necessary to add this with the `rand` feature bubbling in the air-interpreter-data crate, but in the next PR.

mikevoronov: So, in final version `air/Cargo.toml` file is untouched.

monoid: Sorry for the fuss.

crates/air-lib/interpreter-signatures/src/lib.rs
```diff
@@ -168,9 +161,13 @@ impl From<KeyPair> for fluence_keypair::KeyPair {
 }
 
 pub(crate) fn validate_with_key_format<V>(inner: V, key_format: KeyFormat) -> Result<V, KeyError> {
-    match key_format {
-        fluence_keypair::KeyFormat::Ed25519 => Ok(inner),
-        _ => Err(KeyError::AlgorithmNotWhitelisted(key_format)),
+    // this is needed to satisfy forked fluence-keypar crate which contains only single
+    // variant of KeyFormat, this crate should be refactored in the future
```
monoid, line 165: But new algorithm can be added instead.
Or is it intended that fluence_keypair will contain only whilelisted algos?

crates/risc-0/methods/guest/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+[workspace]
+
+[dependencies]
+risc0-zkvm = { version = "0.19.1", default-features = false, features = [ "std" ] }
+zk-aquavm-interface = { path = "../../core" }
+aquavm-air = { path = "../../../../air", version = "0.55.0" }
+
+[patch.crates-io]
+fluence-keypair = { git = "https://github.com/fluencelabs/trust-graph.git", branch = "lean-keypair-2" }
+libp2p-identity = { git = "https://github.com/fluencelabs/rust-libp2p.git", branch = "rand-feature" }
```
monoid, line 19: I believe it is already in the libp2p master.  And the fluence-keypair was an appropriate fix too.

tools/cli/air/src/trace/run/risc0.rs
```diff
@@ -0,0 +1,113 @@
...
+}
+
+fn execute_on_risc0(arguments: AquaVMProvingParameters) -> anyhow::Result<RawAVMOutcome> {
+    use risc0_zkvm::serde::from_slice;
+
+    let env = ExecutorEnv::builder().write(&arguments)?.build()?;
+
+    let prover = default_prover();
+
+    eprintln!("default prover created, proving is started");
```
monoid, line 92: Consider using a logger, e.g. `tracing::info!` (though it may give less readable output).

crates/risc-0/methods/guest/src/main.rs
```diff
@@ -0,0 +1,20 @@
+#![no_main]
```
monoid, line 1: No copyright.

crates/risc-0/methods/build.rs
```diff
@@ -0,0 +1,3 @@
+fn main() {
```
monoid, line 1: No copyright.

crates/risc-0/methods/guest/Cargo.toml
mikevoronov: It is, but there will be another problem b/c Risc-0 supports only Rust 1.69-dev which is too old for libp2p:
```
zk-aquavm: error: package `libp2p-identity v0.2.8` cannot be built because it requires rustc 1.73.0 or newer, while the currently active rustc version is 1.69.0-dev           
z
```

tools/cli/air/src/trace/run/risc0.rs
mikevoronov: These logs are intended b/c it tells that the long-running process has started. If you mean that every single output should be printed with `tracing`, in other parts of the CLI `eprintln` is using, so it's better to keep it consistent.

crates/air-lib/interpreter-signatures/src/lib.rs
monoid: It seems to work with `#[allow(unreachable_patterns)]` enabled locally.

crates/risc-0/methods/guest/Cargo.toml
monoid: 0.2.7 should work:
https://crates.io/crates/libp2p-identity/versions
https://docs.rs/crate/libp2p-identity/0.2.7/features

mikevoronov: There are two problems:
 - as I understand it's not possible to disable or enable features in the patch section if you point to the same version, Rust will try to enable all features that are enabled in all transitive deps of this crate. 
 - currently there is no way in patch to specify a different version than the original dependency, one can only specify the different repo or local path

mikevoronov: For the second an error will be as follows:
```
zk-aquavm: Caused by:
zk-aquavm:   patch for `libp2p-identity` in `https://github.com/rust-lang/crates.io-index` points to the same source, but patches must point to different sources

```

crates/air-lib/interpreter-signatures/src/lib.rs
mikevoronov: Fixed

crates/risc-0/methods/guest/Cargo.toml
mikevoronov: How should it work instead? There is no `libp2p-identity` dep anywhere in the AquaVM repo.

monoid: I don't know how to verify correctly, but I removed the patch section and added
```
libp2p-identity = { version = "=0.2.7", default-features = false }
```
to the dependencies, and the `cargo build` worked.

mikevoronov: Hmm, I understand how version pinning worked here, but I don't think that disabling default-features will really work here b/c the Rust compiler will use all enabled features and won't allow to turned them off...

mikevoronov: We decided to skip this off until RISC-0 supports the newest version of Rust which will make it compilable with the `libp2p-identity` 0.2.8 and then we can implement the `rand` syscall.

monoid (approved):
utACK

PR #682, opened 2023-08-28
feat(execution-engine,interpreter-data)!: insert state for canon join

monoid:
If `canon` is to be executed on another peer and the particle is to be
sent here, current peer inserts canon request-sent-by state to avoid
repeated requests, similar to `call` instruction.

### DIFF
--- a/air/src/execution_step/instructions/canon_utils/mod.rs
+++ b/air/src/execution_step/instructions/canon_utils/mod.rs
@@ -103,7 +103,7 @@ pub(crate) fn handle_unseen_canon(
         exec_ctx.make_subgraph_incomplete();
         exec_ctx.next_peer_pks.push(peer_id);
 
-        trace_ctx.meet_canon_end(CanonResult::send_request_by(
+        trace_ctx.meet_canon_end(CanonResult::request_sent_by(
             exec_ctx.run_parameters.current_peer_id.clone(),
         ));
         Ok(())
--- a/air/src/execution_step/instructions/canon_utils/mod.rs
+++ b/air/src/execution_step/instructions/canon_utils/mod.rs
@@ -65,7 +65,7 @@ pub(crate) fn handle_canon_request_sent_by(
         trace_ctx.meet_canon_end(canon_result);
         Ok(())
     } else {
-        instantiate_canon_stream(epilog, create_canon_stream, peer_id, exec_ctx, trace_ctx)
+        create_canon_stream_for_first_time(epilog, create_canon_stream, peer_id, exec_ctx, trace_ctx)
     }
 }
 
@@ -103,17 +103,15 @@ pub(crate) fn handle_unseen_canon(
         exec_ctx.make_subgraph_incomplete();
         exec_ctx.next_peer_pks.push(peer_id);
 
-        trace_ctx.meet_canon_end(CanonResult::request_sent_by(
-            exec_ctx.run_parameters.current_peer_id.clone(),
-        ));
+        let canon_result = CanonResult::request_sent_by(exec_ctx.run_parameters.current_peer_id.clone());
+        trace_ctx.meet_canon_end(canon_result);
         Ok(())
     } else {
-        instantiate_canon_stream(epilog, create_canon_stream, peer_id, exec_ctx, trace_ctx)
+        create_canon_stream_for_first_time(epilog, create_canon_stream, peer_id, exec_ctx, trace_ctx)
     }
 }
 
-// TODO rename
-fn instantiate_canon_stream(
+fn create_canon_stream_for_first_time(
     epilog: &CanonEpilogClosure<'_>,
     create_canon_stream: &CreateCanonStreamClosure<'_>,
     peer_id: String,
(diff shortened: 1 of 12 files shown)


crates/air-lib/trace-handler/src/merger/canon_merger.rs
```diff
@@ -54,26 +50,34 @@ pub(crate) fn try_merge_next_state_as_canon(data_keeper: &mut DataKeeper) -> Mer
...
+                Ok(prev_canon_result)
+            } else {
+                Err(CanonResultError::incompatible_state(
+                    prev_canon_result,
+                    current_canon_result,
+                ))
+            }
+        }
+        (CanonResult::RequestSentBy(_), CanonResult::Executed(_)) => Ok(current_canon_result),
+        _ => Ok(prev_canon_result),
```
raftedproc, line 81: Do you think that control flow doesn't get here in case of (Executed, RequestSentBy)? If it does this should be an error or am I wrong?

crates/air-lib/interpreter-data/src/executed_state/impls.rs
```diff
@@ -109,8 +109,12 @@ impl ApResult {
 }
 
 impl CanonResult {
-    pub fn new(cid: Rc<CID<CanonResultCidAggregate>>) -> Self {
-        Self(cid)
+    pub fn executed(cid: Rc<CID<CanonResultCidAggregate>>) -> Self {
+        CanonResult::Executed(cid)
+    }
+
+    pub fn send_request_by(peer_id: Rc<String>) -> Self {
```
raftedproc, line 116: Mb rename this to `request_sent_by` to by in sync with the `executed` ctor name?

crates/air-lib/trace-handler/src/merger/canon_merger.rs
```diff
@@ -54,26 +50,34 @@ pub(crate) fn try_merge_next_state_as_canon(data_keeper: &mut DataKeeper) -> Mer
 
 fn prepare_both_canon_result(
```
raftedproc, line 51: Insignificant but IMHO result must be in plural, `prepare_both_canon_results`.

air/tests/test_module/instructions/canon.rs
```diff
@@ -602,3 +617,127 @@ fn canon_map_scalar_with_par() {
 
     assert_eq!(actual_trace, expected_trace);
 }
+
+#[test]
```
raftedproc, line 621: I would check the contents of the resulting canon streams but given the time constraints this can be done later.

monoid: I'm agree with you.

monoid: This test has no resulting canon stream, but we can check intermediate datas.

raftedproc (approved):
utACK

air/src/execution_step/instructions/canon_utils/mod.rs
```diff
@@ -65,15 +102,24 @@ pub(crate) fn handle_unseen_canon(
...
-        //  this canon instruction executes for the first time
-        //  a peer is different from one set in peer_id of a this canon instruction
-        //
-        // the former means that there was no canon associated state in data, and the latter means
-        // that it can't be obtained on this peer, so it's intended not to call meet_canon_end here.
-        return Ok(());
+
+        trace_ctx.meet_canon_end(CanonResult::request_sent_by(
+            exec_ctx.run_parameters.current_peer_id.clone(),
+        ));
```
mikevoronov, line 108: Looks like if the RSB canon result will be created on a separate line, then it should be two not three lines in total.

air/src/execution_step/instructions/canon_utils/mod.rs
```diff
@@ -65,15 +102,24 @@ pub(crate) fn handle_unseen_canon(
...
+        Ok(())
+    } else {
+        instantiate_canon_stream(epilog, create_canon_stream, peer_id, exec_ctx, trace_ctx)
     }
+}
 
+// TODO rename
+fn instantiate_canon_stream(
```
mikevoronov, line 116: What about `create_canon_stream_for_first_time` with a comment "Creates a canon stream for the first time of this script execution".

crates/air-lib/trace-handler/src/merger/canon_merger.rs
```diff
@@ -54,26 +50,34 @@ pub(crate) fn try_merge_next_state_as_canon(data_keeper: &mut DataKeeper) -> Mer
...
+            if prev == cur {
+                Ok(prev_canon_result)
+            } else {
+                Err(CanonResultError::incompatible_state(
+                    prev_canon_result,
+                    current_canon_result,
+                ))
+            }
+        }
+        (CanonResult::RequestSentBy(_), CanonResult::Executed(_)) => Ok(current_canon_result),
```
mikevoronov, line 80: What about the `(Executed, RSB)` option?

mikevoronov: Ok, I see

PR #696, opened 2023-09-11
feat(parser,execution-engine): allow :error: in fail

monoid:
The `(fail :error:)` was previously rejected by parser.  Now it is executed correctly.


air/tests/test_module/instructions/fail.rs
```diff
@@ -97,6 +124,34 @@ fn fail_with_last_error_tetraplets() {
...
+
+    let local_fn_name = "local_fn_name";
+    let script = format!(
+        r#"
+        (xor
+            (xor
+                (call "{local_peer_id}" ("{fallible_service_id}" "{local_fn_name}") [] result_1)
+                (fail :error:)
+            )
+            (call "{local_peer_id}" ("" "") [%last_error%])
```
raftedproc, line 142: What do you think about assigning an `:error:` object to a scalar inside the inner xor and send the scalar as the second call argument to avoid using `%last_error%` that will become obsolete?  
```suggestion
            (call "{local_peer_id}" ("" "") [error_obj_scalar])
```

air/tests/test_module/instructions/fail.rs
```diff
@@ -97,6 +124,34 @@ fn fail_with_last_error_tetraplets() {
...
+    let (host_closure, tetraplet_anchor) = tetraplet_host_function(fallible_call_service(fallible_service_id));
+    let mut vm = create_avm(host_closure, local_peer_id);
+
+    let local_fn_name = "local_fn_name";
+    let script = format!(
+        r#"
+        (xor
+            (xor
+                (call "{local_peer_id}" ("{fallible_service_id}" "{local_fn_name}") [] result_1)
+                (fail :error:)
```
raftedproc, line 140:
```suggestion
                (seq
                  (ap :error: error_object_scalar)
                  (fail :error:)
                )
```

raftedproc (approved):
utACK

mikevoronov (approved):
utACK

air/tests/test_module/instructions/fail.rs
monoid: But it will check value of `:error:` before the `fail`, not after that.
I don't see how to do it correctly, though.

PR #697, opened 2023-09-12
feat(execution-engine): join behavior for canon

monoid:
Implement join behavior for `canon`'s first argument `peer_id`.  It simplifies generating code in Aqua compiler when `peer_id` is defined by a variable defined in a `par` branch.


air/src/execution_step/instructions/canon_utils/mod.rs
```diff
@@ -57,6 +57,8 @@ pub(crate) fn handle_canon_request_sent_by(
     exec_ctx: &mut ExecutionCtx<'_>,
     trace_ctx: &mut TraceHandler,
 ) -> ExecutionResult<()> {
+    // we do not apply join behavior here because if state exists, the variable have been defined;
+    // it cannot become undefined
```
mikevoronov, line 61:
```suggestion
    // it cannot become undefined due to INV-1
```

mikevoronov (approved):
utACK

PR #700, opened 2023-09-18
chore(deps): update polyplets' dependants to drop polyplets 0.4.0

monoid:
As polyplets package has been refactored, it now causes some problems because of version conflict.  Update its dependants so that same version is used everywhere.


monoid:
#640 should be used instead.

PR #701, opened 2023-09-18
fix(execution-engine)!: this adds a join behavior for key and value pair used by ap inserting into a map [fixes VM-337]

raftedproc:
fix(execution-engine)!: this adds a join behavior for key and value pair used by ap inserting into a map [fixes VM-337]


air/tests/test_module/instructions/ap.rs
```diff
@@ -747,3 +747,56 @@ fn ap_canon_stream_map_with_numeric_key_accessor_lambda() {
     ];
     assert_eq!(actual_trace, expected_trace,);
 }
+
+#[test]
+fn ap_map_key_join_behavior() {
+    let vm_1_peer_id = "vm_1_peer_id";
+    let script = format!(
```
monoid, line 754: Why `format!` here and below?

monoid (approved):
utACK

air/tests/test_module/instructions/ap.rs
raftedproc: Fixed

mikevoronov (approved):
utACK

PR #704, opened 2023-09-23
chore: fix typos

omahs:
chore: fix typos

### DIFF
--- a/docs/AIR.md
+++ b/docs/AIR.md
@@ -10,7 +10,7 @@
 - the peer is expected to host Wasm service with the specified `service name`
 - the `service function` is expected to contain the specified function
 - the `arguments list` is given to the function and may be empty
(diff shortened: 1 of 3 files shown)


docs/AIR.md
```diff
@@ -87,7 +87,7 @@ Examples:
 (mismatch <variable> <variable> <instruction>)
 ```
 
-- executes the instruction iff variables are equal/notequal
+- executes the instruction if variables are equal/notequal
```
monoid, line 90: Hi! Thanks for the corrections!
I think `iff` means here "if and only if", quite popular abbreviation. It may be the case above too.

omahs: Ohh sorry about this, I changed it back.

mikevoronov (approved):
utACK.

Thank you for the corrections!

PR #705, opened 2023-09-25
fix(execution-engine): ExecutionCtx fmt now prints stream maps out

raftedproc:
This is a forgotten change for the maps patch.


monoid (approved):
utACK

PR #706, opened 2023-09-25
fix(execution-engine)!: map with a lens now returns an appropriate tetraplet [fixes VM-331]

raftedproc:
Map with a lens now returns an appropriate tetraplet that is described in both map spec and vm-331


air/src/execution_step/lambda_applier/applier.rs
```diff
@@ -104,13 +111,62 @@ fn select_by_path_from_stream<'value>(
...
+    } else {
+        let SecurityTetraplet {
+            peer_pk,
+            service_id,
+            function_name,
+            json_path,
+        } = tetraplet.as_ref();
+
+        let json_path_suffix = body.iter().fold("".to_string(), |acc, va| acc + &va.to_string());
+        let json_path = json_path.to_string() + &prefix.to_string() + &json_path_suffix;
```
mikevoronov, line 154: you can use `add_lambda` method

raftedproc: But I create a new tetraplet using the constructed path. Should I use add_lambda() in this case?

air/src/execution_step/lambda_applier/applier.rs
```diff
@@ -78,22 +86,9 @@ fn select_by_path_from_stream<'value>(
...
-        }
-        ValueAccessor::FieldAccessByScalar { scalar_name } => {
-            let scalar = exec_ctx.scalars.get_value(scalar_name)?;
-            lambda_to_execution_error!(try_scalar_ref_as_idx(scalar))?
-        }
-        ValueAccessor::Error => unreachable!("should not execute if parsing succeeded. QED."),
-    };
-    let idx = idx as usize;
     let stream_size = stream.len();
+    let (idx, body) = exctract_idx(lambda, exec_ctx)?;
```
monoid, line 90: Should be "extract".

air/src/execution_step/lambda_applier/applier.rs
```diff
@@ -104,13 +99,43 @@ fn select_by_path_from_stream<'value>(
...
+) -> ExecutionResult<MapLensResult<'value>> {
+    let stream_size = stream.len();
+    let (idx, body) = exctract_idx(lambda, exec_ctx)?;
+
+    let (value, tetraplet) = lambda_to_execution_error!(stream
+        .peekable()
+        .nth(idx)
+        .ok_or(LambdaError::CanonStreamNotHaveEnoughValues { stream_size, idx }))?;
+
+    // csm.$.key.[0] case
```
monoid, line 115: The comment should be moved inside the "then" block.

air/src/execution_step/lambda_applier/applier.rs
```diff
@@ -104,13 +99,43 @@ fn select_by_path_from_stream<'value>(
...
+        MapLensResult::from_cow(result, tetraplet)
+    } else {
+        // csm.$.key.[0].attribute case
+        let result = select_by_path_from_scalar(value, body.iter(), exec_ctx)?;
+
+        let json_path_suffix = body.iter().fold("".to_string(), |acc, va| acc + "." + &va.to_string());
+        let prefix_with_path = true;
+        let tetraplet = update_tetraplet_with_path(&tetraplet, &json_path_suffix, prefix_with_path);
+
+        MapLensResult::from_cow(result, tetraplet)
```
monoid, line 127: `...::from_cow` call is common for both branches.  Extracting common code reduces code complexity.

air/src/execution_step/lambda_applier/applier.rs
```diff
@@ -132,31 +157,74 @@ fn select_by_path_from_canon_map<'value>(
...
+            let scalar = exec_ctx.scalars.get_value(scalar_name)?;
+            lambda_to_execution_error!(try_scalar_ref_as_idx(scalar))?
+        }
+        ValueAccessor::Error => unreachable!("should not execute if parsing succeeded. QED."),
+    };
+    Ok((idx as usize, body))
+}
+
+fn update_tetraplet_with_path(
+    original_tetraplet: &RcSecurityTetraplet,
```
monoid, line 208: It can be just `&SecurityTetraplet`.

air/src/execution_step/lambda_applier/applier.rs
```diff
@@ -132,31 +157,74 @@ fn select_by_path_from_canon_map<'value>(
...
+        ValueAccessor::FieldAccessByScalar { scalar_name } => {
+            let scalar = exec_ctx.scalars.get_value(scalar_name)?;
+            lambda_to_execution_error!(try_scalar_ref_as_idx(scalar))?
+        }
+        ValueAccessor::Error => unreachable!("should not execute if parsing succeeded. QED."),
+    };
+    Ok((idx as usize, body))
+}
+
+fn update_tetraplet_with_path(
```
monoid, line 207: Perhaps, it should belong to `SecurityTetraplet` type.  It is in foreign crate now, but it worth adding a TODO comment.

air/src/execution_step/lambda_applier/applier.rs
```diff
@@ -132,31 +157,74 @@ fn select_by_path_from_canon_map<'value>(
...
+    original_tetraplet: &RcSecurityTetraplet,
+    original_path: &impl Display,
```
monoid, line 209: I think you should require `ToString` here.  Some types may fail to implement `Display`, but have `ToString` implemented.

monoid: Perhaps, `AsRef<str>` even better, as it avoids allocation.

air/src/execution_step/lambda_applier/applier.rs
```diff
@@ -231,3 +303,22 @@ impl<'value> LambdaResult<'value> {
         }
     }
 }
+
+impl<'value> MapLensResult<'value> {
+    fn from_cow(result: Cow<'value, JValue>, tetraplet: RcSecurityTetraplet) -> Self {
```
monoid, line 308: I'd call the method so simple just `new`.

air/src/execution_step/lambda_applier/applier.rs
```diff
@@ -132,31 +157,74 @@ fn select_by_path_from_canon_map<'value>(
...
+        json_path,
+    } = original_tetraplet.as_ref();
+
+    let json_path = if prefix_with_path {
+        json_path.to_string() + &original_path.to_string()
+    } else {
+        original_path.to_string()
+    };
+
+    SecurityTetraplet::new(peer_pk, service_id, function_name, json_path).into()
```
monoid, line 225: BTW, struct update syntax would make this function shorter.  But I am ambivalent to it.

air/src/execution_step/lambda_applier/applier.rs
```diff
@@ -104,13 +99,43 @@ fn select_by_path_from_stream<'value>(
...
+
+    // csm.$.key.[0] case
+    let select_result = if body.is_empty() {
+        let result = Cow::Borrowed(value);
+        MapLensResult::from_cow(result, tetraplet)
+    } else {
+        // csm.$.key.[0].attribute case
+        let result = select_by_path_from_scalar(value, body.iter(), exec_ctx)?;
+
+        let json_path_suffix = body.iter().fold("".to_string(), |acc, va| acc + "." + &va.to_string());
```
monoid, line 123: BTW, the lambda expression can be rewritten with `format!`, and it will be more efficient than the string addition.

air/src/execution_step/lambda_applier/applier.rs
```diff
@@ -132,31 +157,74 @@ fn select_by_path_from_canon_map<'value>(
...
-                json_path,
-            } = canon_map.tetraplet().as_ref();
-            let json_path = json_path.to_string() + &prefix.to_string();
-
-            let tetraplet = SecurityTetraplet::new(peer_pk, service_id, function_name, json_path).into();
-            let value = CanonStream::new(vec![], tetraplet).as_jvalue();
-            Cow::Owned(value)
+            // csm.$.non_existing_key case
+            let prefix_with_path = false;
+            let tetraplet = update_tetraplet_with_path(canon_map.tetraplet(), original_lambda, prefix_with_path);
```
monoid, line 176: `update_tetraplet_with_path` should be simplified and split into two functions: `update_...` witch can use `SecurityTetraplet::add_path`, and `replace_...` that may use struct update.  The flag seem to make code less clear.

air/src/execution_step/lambda_applier/applier.rs
monoid: I've been tricked by the shadowing of the `tetraplet` variable.  Such shadowing makes code harder to read.

air/src/execution_step/lambda_applier/applier.rs
raftedproc: According with my research format! won't reduce the number of intermediate String results but Vec::join() will.
```
body.iter().map(ToString::to_string).collect::<Vec<_>>().join(".");
```

air/src/execution_step/lambda_applier/applier.rs
raftedproc: Can't use AsRef, b/c LambdaAst can't be used as a str slice.

air/src/execution_step/lambda_applier/applier.rs
raftedproc: This breaks the naming schema used just above in impl of LambdaResult.

air/src/execution_step/lambda_applier/applier.rs
monoid: First, `format!` will store everything into one buffer (yep, recursively).  You implementation allocates a new string for each element.

Second, `format!` is idiomatic way of concatenating string, especially if some of them are static: https://rust-unofficial.github.io/patterns/idioms/concat-format.html

And I was talking about the lambda expression, not whole loop.  BTW, you should use [`Itertools::format`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.format) with `format_args!`.

air/src/execution_step/lambda_applier/applier.rs
```diff
@@ -104,13 +98,44 @@ fn select_by_path_from_stream<'value>(
     Ok(select_result)
 }
 
+fn select_by_path_from_canon_map_stream<'value>(
+    stream: impl ExactSizeIterator<Item = (&'value JValue, RcSecurityTetraplet)> + 'value,
+    lambda: &NonEmpty<ValueAccessor<'_>>,
+    exec_ctx: &ExecutionCtx<'_>,
+) -> ExecutionResult<MapLensResult<'value>> {
+    let stream_size = stream.len();
+    let (idx, body) = extract_idx(lambda, exec_ctx)?;
```
mikevoronov, line 107: I'd rename it into 'split_to_idx'

air/src/execution_step/lambda_applier/applier.rs
```diff
@@ -104,13 +98,44 @@ fn select_by_path_from_stream<'value>(
...
+    let select_result = if body.is_empty() {
+        // csm.$.key.[0] case
+        let result = Cow::Borrowed(value);
+        MapLensResult::from_cow(result, tetraplet)
+    } else {
+        // csm.$.key.[0].attribute case
+        let result = select_by_path_from_scalar(value, body.iter(), exec_ctx)?;
+
+        let joined = body.iter().map(ToString::to_string).collect::<Vec<_>>().join(".");
+        let json_path_suffix = format!(".{}", joined);
```
mikevoronov, line 123: It's an abstraction leakage, no one should know about how lens is represented internally (e.g. it's united with a dot). One should use `add_lambda` instead.

air/src/execution_step/lambda_applier/applier.rs
raftedproc: Done

air/src/execution_step/lambda_applier/applier.rs
mikevoronov: We decided to do it in the next PR.

mikevoronov (approved):
utACK

PR #708, opened 2023-09-27
fix(execution-engine)!: canon map lens access now returns a correct number of kvpairs in a canon stream

raftedproc:
Previously `#%canon_map.$.[key]` returned FWW value from an underlying stream.


mikevoronov (approved):
utACK

PR #712, opened 2023-10-04
chore(benches): Update bench data

monoid:
Some benchmarks were removed because they are not compatible with the stream length limit.  Others have their version updated.


ValeryAntopol (approved):
utACK

mikevoronov (approved):
utACK

raftedproc (approved):
utACK

PR #713, opened 2023-10-12
feat(execution-engine)!: `fail :error:` now bubbles the original error up [fixes VM-342]

raftedproc:
AquaVM wraps an original error with Catchable::UserError for every xor with fail used. This patch bubbles up an original error with the corresponding error_object.

### DIFF
--- a/air/src/execution_step/errors/catchable_errors.rs
+++ b/air/src/execution_step/errors/catchable_errors.rs
@@ -32,7 +32,7 @@ use std::rc::Rc;
 /// Catchable errors arisen during AIR script execution. Catchable here means that these errors
 /// could be handled by a xor instruction and their error_code could be used in a match
 /// instruction.
-#[derive(ThisError, EnumDiscriminants, Debug)]
+#[derive(ThisError, EnumDiscriminants, Debug, Clone)]
 #[strum_discriminants(derive(EnumIter))]
 pub enum CatchableError {
     /// An error is occurred while calling local service via call_service.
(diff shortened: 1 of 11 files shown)


air/src/execution_step/errors/catchable_errors.rs
```diff
@@ -32,7 +32,7 @@ use std::rc::Rc;
 /// Catchable errors arisen during AIR script execution. Catchable here means that these errors
 /// could be handled by a xor instruction and their error_code could be used in a match
 /// instruction.
-#[derive(ThisError, EnumDiscriminants, Debug)]
+#[derive(ThisError, EnumDiscriminants, Debug, Clone)]
```
monoid, line 35: Can we keep this value in a `Rc` instead?

air/src/execution_step/execution_context/context.rs
```diff
@@ -193,8 +193,14 @@ impl ExecutionCtx<'_> {
             None
         };
 
+        let original_error_object = self.error_descriptor.original_error_object().clone();
```
monoid, line 196: You clone it, but later pass it by reference only.

raftedproc:
Closing b/c rewrote the patch substrantially.

air/src/execution_step/errors/catchable_errors.rs
raftedproc: There is ExecutionError instance in Xor::execute that must be cloned before send down anyway.

PR #718, opened 2023-10-16
feat(execution-engine,test-utils,interpreter-data,interpreter-cid)!: Rc into CID

monoid:
* Hide `Rc` into `CID` type, making it cheap to clone.
* Introduce `CidRef` type that abstracts on `CID`'s inner type.

This change makes memory more optimal (single allocation vs two allocations) and makes it easier to change CID's internal representation from string to binary.


mikevoronov (approved):
utACK

PR #723, opened 2023-10-16
chore(testing-framework): a new behavior to print service call parameters/result out to console

raftedproc:
This patch introduces new test framework behavior to simplify AIR tests troubleshooting.


crates/testing-framework/src/asserts/behavior.rs
```diff
@@ -66,7 +68,8 @@ impl Behavior {
         use Behavior::*;
 
         match self {
-            Echo => {
+            Echo => echo_call_service()(params),
+            DbgEcho => {
                 println!("echo_call_service() {:#?}", params);
```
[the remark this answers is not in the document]
monoid: You don't need to clone anything as `dbg!` just returns its argument.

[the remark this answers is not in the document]
monoid: You also may use `dbg!(&var);`.

crates/testing-framework/src/asserts/mod.rs
```diff
@@ -54,6 +54,9 @@ pub enum ServiceDefinition {
     /// Some known service by name: "echo", "unit" (more to follow).
     #[strum_discriminants(strum(serialize = "behaviour"))]
     Behaviour(Behavior),
+    /// Same services as defined by the enum element above with dbg! applied to the arguments.
+    #[strum_discriminants(strum(serialize = "dbg_behaviour"))]
```
monoid, line 58: I think calling it just `dbg` is more ergonomic.

air/tests/test_module/features/errors/error.rs
```diff
@@ -29,9 +29,9 @@ fn fail_with_rebubble_error() {
                 (match 1 2 (null) )
                 (fail :error:)
             )
-            (call "peer_id" ("m" "f1") [:error:] scalar1) ; behaviour = echo
+            (call "peer_id" ("m" "f1") [:error:] scalar1) ; dbg_behaviour = echo
         )
-        (call "peer_id" ("m" "f2") [:error:] scalar2) ; behaviour = echo
+        (call "peer_id" ("m" "f2") [:error:] scalar2) ; dbg_behaviour = echo
```
monoid, line 34: There are dedicated tests in the `crates/testing-framework/src/execution/mod.rs`.

crates/testing-framework/src/asserts/mod.rs
```diff
@@ -103,6 +106,7 @@ impl ServiceDefinition {
                 call_map,
             } => call_seq_error(call_number_seq, call_map),
             ServiceDefinition::Behaviour(name) => name.call(params),
+            ServiceDefinition::DbgBehaviour(name) => name.call(dbg!(params)),
```
monoid, line 109: Wouldn't be nice to dbg result too?

crates/testing-framework/src/asserts/mod.rs
raftedproc: But `dbg` tells nothing regarding the actual effect that happens.

mikevoronov: I also vote for `dbg_behaviour`, it makes more sense for people who not familiar with Rust-specific `dbg` function.

crates/testing-framework/src/asserts/mod.rs
mikevoronov: Agreed

mikevoronov (approved):
utACK

PR #731, opened 2023-10-23
feat(air-interpreter)!: enable signature generation and verification features

monoid:
Make `gen_signatures` and `check_signature` features defaults.


mikevoronov (approved):
utACK

PR #734, opened 2023-10-25
feat(interpreter-data)!: allow only deterministic signature algorithms

monoid:
Reject algorithms that require random number generators for generating safe signatures.  Both signing keys and other peers' signatures are checked.


mikevoronov (approved):
utACK

PR #735, opened 2023-10-25
chore(benchmarking): data generators for a set of map-related benchmarks [fixes VM-345]

raftedproc:
This adds data generators for map operations benchmarks.


junk/gen-bench-data/src/main.rs
```diff
@@ -59,41 +76,64 @@ fn save_data(dest_dir: &Path, data: Data) -> Result<(), Box<dyn std::error::Erro
...
+    filename: &str,
+    data: Option<impl AsRef<[u8]>>,
+) -> Result<(), Box<dyn std::error::Error>> {
     use std::fs::*;
     use std::io::prelude::*;
 
     let mut dest_dir = dest_dir.to_owned();
     dest_dir.push(filename);
 
     let mut f = File::create(&dest_dir)?;
```
mikevoronov, line 127: Is it intended to create an empty file in case of no `data`? Otherwise it's better to first check for data presence and only then create a file.

junk/gen-bench-data/src/main.rs
```diff
@@ -232,3 +275,377 @@ fn multiple_sigs(size: usize) -> Data {
...
+        TestRunParameters::from_init_peer_id("init_peer_id").with_particle_id(PARTICLE_ID),
+        vec![],
+        vec![],
+        &air_script,
+    )
+    .unwrap();
+
+    let prev_res = exec.execute_one("init_peer_id").unwrap();
+
+    // let cur_res: RawAVMOutcome = exec.execute_one("other_peer_id").unwrap();
```
mikevoronov, line 427: looks obsolete

junk/gen-bench-data/src/main.rs
```diff
@@ -59,41 +76,64 @@ fn save_data(dest_dir: &Path, data: Data) -> Result<(), Box<dyn std::error::Erro
...
-    f.write_all(data.as_ref())?;
+    if let Some(data) = data {
+        f.write_all(data.as_ref())?;
+    }
 
     Ok(())
 }
 
 #[derive(Debug, Default)]
+#[allow(dead_code)]
```
mikevoronov, line 136: Is it needed?

mikevoronov (commented):
There are multiple new line at the end of air script files are missed.

junk/gen-bench-data/src/main.rs
raftedproc: call_results is not used and my linter complains about this. I don't want to remove this attribute b/c it will be useful at some point.f

.gitmodules
monoid, line 1: Hey, don't remove this file.

junk/gen-bench-data/gen_benchmark_data.sh
```diff
@@ -7,14 +7,20 @@ set -e
 echo "Pre-build a binary..." >&2
 cargo build --quiet --release
 
-for bench in multiple-cids10 \
-             multiple-peers8 \
-             multiple-sigs30 \
-             dashboard network-explore; do
+for bench in canon-map-key-by-lens \
+            canon-map-key-element-by-lens \
```
monoid, line 11: 1. I think it worth adding a similar bench for simple stream as a baseline.
2. I think it worth explicitly mentioning a problem/data size in the bench name (we may want to have same bench instances of different sizes).

raftedproc: Agree but IMHO it should be done with a separate patch.

monoid: Problem size should be added as early as possible.

air/benches/data/anomaly_big.json
monoid, line 1: The files in `air/benches/data` are used in the `air/benches` benchmarks (ordinary Criterion benchmarks). They are broken, of course.  But I'm not sure if removing them is OK.

junk/gen-bench-data/src/cid_benchmarking/big_values_data_prev_data.json
monoid, line 1: This and the other file are not needed anymore.

junk/cidify/Cargo.lock
monoid, line 1: Hm, this file was added by mistake.

junk/gen-bench-data/src/populate_map.air.tmpl
```diff
@@ -0,0 +1,25 @@
+(new %map
+  (seq
+    (seq
+      (call "init_peer_id" ("" "init") [] data) ; ok = [{data}]
+      (seq
+        (canon "other_peer_id" %map #%switch_node1)
+        (seq
+          (null)
```
monoid, line 8: Is it really needed?

monoid (approved):
Tested ACK

PR #750, opened 2023-11-01
chore(docs): an additional doc on fold

raftedproc:


docs/fold.md
```diff
@@ -0,0 +1,82 @@
+## AIR: Fold instruction
+Here is the `fold` syntax.
+
+`(fold <iterable> <iterator> <instruction> [<last_instruction>])`
+
```
mikevoronov, line 5: Let add here that `Iterable` can be a stream, map, canonicalized stream, canonicalized map or scalar array.

docs/fold.md
```diff
@@ -0,0 +1,82 @@
...
+				(ap 42 $data)
+					...
+			)
+		)
+	)
+	;; here is an implicit (null)
+)
+```
+
+Fold internally is a vector of vectors where inner vectors are called generations. In the previous case ap adds a new generation into `fold`.
```
mikevoronov, line 66:
```suggestion
A stream internally is a vector of vectors where inner vectors are called generations. In the previous case ap adds a new generation into `fold`.
```

docs/fold.md
```diff
@@ -0,0 +1,82 @@
+## AIR: Fold instruction
+Here is the `fold` syntax.
+
+`(fold <iterable> <iterator> <instruction> [<last_instruction>])`
+
+`last_instruction` is:
+- an implicit `null` for scalar-based iterables
+- an implicit `never` for stream-based iterables
```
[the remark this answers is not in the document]
raftedproc: Fixed

docs/fold.md
```diff
@@ -0,0 +1,82 @@
...
+(fold data value
+	(seq
+	  (ap ("somesome-lengthy-and-scary-key" value) %map)
+	  (next value)
+	)
+	;; here is an implicit (null)
+)
+```
+
+Here is an operational semantics of fold is if data is `[1 2 3]`. Please note that the recursion produced by `next` expressed in recursion of AquaVM execution engine function calls. Putting this differently the depth of the recursion depends on the size of the iterable.
```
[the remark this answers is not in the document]
raftedproc: Fixed

docs/fold.md
raftedproc: Fixed

docs/fold.md
raftedproc: Fixed

docs/fold.md
```diff
@@ -0,0 +1,84 @@
+## AIR: Fold instruction
+Here is the `fold` syntax.
+
+`(fold <iterable> <iterator> <instruction> [<last_instruction>])`
```
InversionSpaces, line 4: How does `fold` handle errors in its `instruction`?

docs/fold.md
```diff
@@ -0,0 +1,84 @@
...
+
+If `last_instruction` is not set, then it's assumed:
+ - `null` in case of `(fold (seq ...`
+ - `never` in case of `(fold (par ...`
+
+```clojure
+(fold iterable iterator
+	(seq
+	  (ap ("key" iterator) %map)
+	  (next iterator)
```
InversionSpaces, line 16: Would be nice to provide an example with `next` not as last instruction in `isntruction` also

InversionSpaces: And elaborate on what happens if `next` is **not** called inside `instruction`

docs/fold.md
```diff
@@ -0,0 +1,84 @@
...
+
+A stream internally is a vector of vectors where inner vectors are called generations. In the previous case ap adds a new generation into `fold`.
+
+Here is a high level overview how `fold` handles multiple generations of an iterable.
+
+```clojure
+fold
+	run fold body for generation 1 values
+		run last instruction
+	run fold body for generation 2 values
```
InversionSpaces, line 76: Runs of previous generations do not affect next generations in any way? 
What if previous generation didn't call `next`? Next generation will still be executed?

docs/fold.md
```diff
@@ -0,0 +1,84 @@
...
+		run last instruction
+	...
+	run fold body for generation N values
+		run last instruction
+```
+
+There is a property of `fold` completeness. The invariant for this property is as follows, if fold finishes a run for at least one generation the fold is marked as complete. This property belongs to AquaVM execution engine and it is not explicitly sent out in a particle.
```
InversionSpaces, line 83: But AquaVM tries to run other generations still?

docs/fold.md
mikevoronov: > Runs of previous generations do not affect next generations in any way?
It doesn't

What if previous generation didn't call next? Next generation will still be executed?
Yep, it doesn't depend on whether next was called or not.

docs/fold.md
mikevoronov: Only fold over stream influences error, actually it supresses them, otherwise it'd be not deterministic. Fold over scalar and canonicalized maps and streams doesn't influence errors anyhow, the best analogy is the provided one with a tree, so it behaves exactly like in this tree representation.

docs/fold.md
raftedproc: Good point

docs/fold.md
raftedproc: It does.

docs/fold.md
```diff
@@ -0,0 +1,126 @@
...
+(fold iterable iterator
+	(seq
+	  (ap ("key" iterator) %map)
+	  (next iterator)
+	)
+	;; here is an implicit (null)
+)
+```
+
+This example demonstrates the case when `next` is not the last instruction. This feature is supported with `fold`` over scalar only.
```
InversionSpaces, line 23:
```suggestion
This example demonstrates the case when `next` is not the last instruction. This feature is supported with `fold` over scalar only.
```

docs/fold.md
```diff
@@ -0,0 +1,126 @@
...
+(fold iterable iterator
+	(seq
+	  (next iterator)
+	  (ap ("key" iterator) %map)
+	)
+	;; here is an implicit (null)
+)
+```
+
+Fold body can have no `next`. The body will be executed only once in this case.
```
InversionSpaces, line 34: I think it is important to emphasize that this is true for `fold` over scalars and canons only

docs/fold.md
```diff
@@ -0,0 +1,126 @@
...
+fold
+	run fold body for generation 1 values
+		run last instruction
+	run fold body for generation 2 values
+		run last instruction
+	...
+	run fold body for generation N values
+		run last instruction
+```
+There are some subtle details about fold processing over a number of generations, namely:
```
InversionSpaces, line 103:
```suggestion
There are some subtle details about `fold` processing over a number of generations, namely:
```

docs/fold.md
```diff
@@ -0,0 +1,126 @@
...
+- runs of previous generations do not affect next generations in any way
+- if previous generation doesn't call `next` AquaVM executes `fold` for the next generation.
```
InversionSpaces, line 105:
```suggestion
- even if previous generation doesn't call `next` AquaVM executes `fold` for the next generation.
```

docs/fold.md
raftedproc: Fixed

docs/fold.md
```diff
@@ -0,0 +1,131 @@
...
+			(seq ;; next
+				(ap ("key" iterator) %map)
+				;; an implicit (null)
+			)
+		)
+	)
+)
+```
+
+There is so-called recursive fold which populates iterable stream `fold` is iterating over.
```
InversionSpaces, line 64: Why is this example not in the next "`fold` for stream" section?

docs/fold.md
```diff
@@ -0,0 +1,131 @@
...
+		run last instruction
+	run fold body for generation 2 values
+		run last instruction
+	...
+	run fold body for generation N values
+		run last instruction
+```
+There are some subtle details about `fold` processing over a number of generations, namely:
+- runs of previous generations do not affect next generations in any way
+- even if previous generation doesn't call `next` AquaVM executes `fold` for the next generation.
```
InversionSpaces, line 110: But if `next` is not called in one generation, other elements in this generation are not processed?

docs/fold.md
```diff
@@ -0,0 +1,131 @@
...
+
+There is a property of `fold` completeness. The invariant for this property is as follows, if fold finishes a run for at least one generation the fold is marked as complete. This property belongs to AquaVM execution engine and it is not explicitly sent out in a particle.
```
InversionSpaces, line 112:
```suggestion
There is a property of `fold` completeness. The invariant for this property is as follows, if fold finishes a run for at least one generation the `fold` is marked as complete. This property belongs to AquaVM execution engine and it is not explicitly sent out in a particle.
```

docs/fold.md
```diff
@@ -0,0 +1,131 @@
...
+Please note that fold body that triggers a node change events, e.g. `call`,`canon` destined for a node other than the current, forces the executiont to migrate to the destined node.
```
InversionSpaces, line 113:
```suggestion
Please note that `fold` body that triggers a node change events, e.g. `call`, `canon` destined for a node other than the current, forces the execution to migrate to the destined node.
```

docs/fold.md
```diff
@@ -0,0 +1,131 @@
...
+Fold body can have no `next`. The body will be executed only once in this case.
+
+```clojure
+(fold iterable iterator
+	  (ap ("key" iterator) %map)
+	;; here is an implicit (null)
+)
+```
+
+Let's consider the fold's operational semantics if iterable is the following scalar array `[1 2 3]`. Note that the recursion produced by `next` expressed in recursion of AquaVM execution engine function calls. Putting this differently the depth of the recursion depends on the size of the iterable.
```
InversionSpaces, line 47:
```suggestion
Let's consider the `fold`'s operational semantics if iterable is the following scalar array `[1 2 3]`. Note that the recursion produced by `next` expressed in recursion of AquaVM execution engine function calls. Putting this differently the depth of the recursion depends on the size of the iterable.
```

docs/fold.md
```diff
@@ -0,0 +1,131 @@
...
+(fold iterable iterator
+	(seq
+	  (next iterator)
+	  (ap ("key" iterator) %map)
+	)
+	;; here is an implicit (null)
+)
+```
+
+Fold body can have no `next`. The body will be executed only once in this case.
```
InversionSpaces, line 38:
```suggestion
The `fold` body can have no `next`. The body will be executed only once in this case.
```

docs/fold.md
DieMyst: can we dump answers into documentation too?

DieMyst:
There should be section about error handling inside `fold` instruction

mikevoronov (approved):
utACK

raftedproc:
> There should be section about error handling inside `fold` instruction

What exactly are we talking about?
:error: access or what?

PR #753, opened 2023-11-03
fix(air-interpreter): enable multiple invoke_tracing calls for air-interpreter

raftedproc:
A hacky way to prevent multiple initialization of a tracing subscruber.
This is a prerequisite to add benchmarking tests for nox.

### DIFF
--- a/air-interpreter/src/marine.rs
+++ b/air-interpreter/src/marine.rs
@@ -37,6 +37,8 @@ use marine_rs_sdk::module_manifest;
 
 module_manifest!();
 
+static mut TRACING_IS_INITED: bool = false;
+
 pub fn main() {
     logger::init_logger(None);
 }
@@ -63,7 +65,12 @@ pub fn invoke_tracing(
     tracing_params: String,
     tracing_output_mode: u8,
 ) -> InterpreterOutcome {
-    logger::init_tracing(tracing_params, tracing_output_mode);
+    if unsafe { !TRACING_IS_INITED } {
+        logger::init_tracing(tracing_params, tracing_output_mode);
+        unsafe {
+            TRACING_IS_INITED = true;
+        }
+    }
     execute_air(air, prev_data, data, params, call_results)
 }
 
--- a/air-interpreter/src/marine.rs
+++ b/air-interpreter/src/marine.rs
@@ -37,8 +37,6 @@ use marine_rs_sdk::module_manifest;
 
 module_manifest!();
 
-static mut TRACING_IS_INITED: bool = false;
-
 pub fn main() {
     logger::init_logger(None);
 }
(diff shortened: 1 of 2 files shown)


mikevoronov (approved):
utACK

air-interpreter/src/marine.rs
```diff
@@ -63,7 +65,12 @@ pub fn invoke_tracing(
     tracing_params: String,
     tracing_output_mode: u8,
 ) -> InterpreterOutcome {
-    logger::init_tracing(tracing_params, tracing_output_mode);
+    if unsafe { !TRACING_IS_INITED } {
```
monoid, line 68: I've looked at the tracing_subscriber docs.  It seems we can use two alternative approaches:
1. `init_tracing` should use `try_init`.  However, it returns opaque error object, and we cannot tell if it is existing object or some other error.
2. Run `execute_air` inside `with_default(...)`.  It seems to be perfect solution as it allows different tracing options on each invocation.

See https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/fn.fmt.html for examples for both options.

air-interpreter/src/marine.rs
```diff
@@ -65,13 +63,21 @@ pub fn invoke_tracing(
...
+        .with_span_events(FmtSpan::ENTER | FmtSpan::CLOSE)
+        .with_writer(std::io::stderr);
+
+    if logger::json_output_mode(tracing_output_mode) {
+        let subscriber = builder.json().finish();
+        tracing::subscriber::with_default(subscriber, || execute_air(air, prev_data, data, params, call_results))
+    } else {
+        // Human-readable output.
+        let subscriber = builder.finish();
+        tracing::subscriber::with_default(subscriber, || execute_air(air, prev_data, data, params, call_results))
```
monoid, line 79: This code looks WTF.  I know that it has a reason: `subscriber` has different type in both branches.  It can be simplified with `Box<dyn ...>` or with some feature of `tracing`: `tracing::subscriber::with_default` simply creates `Dispatch` instance inside, and `Dispatch` type is not parametrized by subscriber type.  So:

```rust
    let dispatch = if logger::json_output_mode(tracing_output_mode) {
        let subscriber = builder.json().finish();
        Dispatch::new(subscriber)
    } else {
        // human-readable output
        let subscriber = builder.finish();  // n.b.: this `subscriber` has different type
        Dispatch::new(subscriber)
    };
    tracing::dispatcher::with_default(&dispatch, || execute_air(air, prev_data, data, params, call_results))
```

raftedproc: This looks like a leaked abstraction to me. You are trying to use an entity that was wrapped for a reason.

raftedproc: Changed though.

monoid: How do you tell it is a leaking abstraction, but not a more low-level API?

raftedproc: There is a section on dispatcher module in the tracing docs so I might be wrong classifying Dispatcher as an internal entity.

raftedproc: Anything else to tell about this PR?

monoid (approved):
utACK

PR #756, opened 2023-11-21
feat(validator): add after next instructions check

raftedproc:
This check is supposed to block fold blocks for stream and stream-based structs to avoid call stack implosion.


air/tests/test_module/features/data_merging/data_merge.rs
```diff
@@ -219,6 +219,7 @@ fn stream_merge() {
 }
 
 #[test]
+#[ignore]
```
monoid, line 222: It worth clarifying this change (as well as below).

monoid: Otherwise it will certainly be ignored forever.

raftedproc: Mike had suggested this.

PR #759, opened 2023-11-27
chore(benchmarks): call-requests and call-results benchmarks

monoid:


raftedproc (commented):
Do you need to update the global Cargo.lock in the context of this patch?

monoid:
@raftedproc it is updated because someone forgot to update it in the previous commits.

benches/PERFORMANCE.json
```diff
@@ -3792,255 +3874,667 @@
...
-          "58.938 MiB",
-          "58.938 MiB"
+          "59.000 MiB",
+          "59.000 MiB"
         ],
         "stats": {
           "air::runner::execute_air": {
             "common_prefix": "air",
-            "duration": "12.11ms",
+            "duration": "17.72ms",
```
raftedproc, line 3912: I am curious if this is a degradation and something we expect?

raftedproc: Should be signatures processing overhead?

monoid: CID verification.

PR #760, opened 2023-11-28
feat(parser): Instructions after next validator check

raftedproc:
This check prohibits instructions in the `fold` body after the corresponding `next` for stream/stream map.

### DIFF
--- a/crates/air-lib/air-parser/src/parser/span.rs
+++ b/crates/air-lib/air-parser/src/parser/span.rs
@@ -42,8 +42,8 @@ impl Span {
 
     // This presumes that a `span` that begins in `self` can not overlap
     // and at the same time can not get outside the `self`.
-    pub fn contains_or_equal_span(&self, span: Self) -> bool {
-        self.left <= span.left && span.right <= self.right
+    pub fn contains_or_equal_span(&self, target_span: Self) -> bool {
+        self.left <= target_span.left && target_span.right <= self.right
     }
 }
 
--- a/crates/air-lib/air-parser/src/parser/span.rs
+++ b/crates/air-lib/air-parser/src/parser/span.rs
@@ -39,6 +39,12 @@ impl Span {
     pub fn contains_span(&self, span: Self) -> bool {
         self.contains_position(span.left) && self.contains_position(span.right)
     }
+
+    // This presumes that a `span` that begins in `self` can not overlap
+    // and at the same time can not get outside the `self`.
+    pub fn contains_or_equal_span(&self, span: Self) -> bool {
+        self.left <= span.left && span.right <= self.right
+    }
 }
 
 impl From<Range<AirPos>> for Span {
--- a/crates/air-lib/air-parser/src/parser/validator.rs
+++ b/crates/air-lib/air-parser/src/parser/validator.rs
@@ -595,16 +595,10 @@ impl<'i> ValidatorErrorBuilder<'i> {
                 {
                     {
                         let after_next_right_pos = fold_span.right;
-                        if self
-                            .validator
-                            .other_instrs_spans
-                            .iter()
-                            .find(|instr_span| {
-                                after_next_left_pos < instr_span.left
-                                    && instr_span.right < after_next_right_pos
-                            })
-                            .is_some()
-                        {
+                        if self.validator.other_instrs_spans.iter().any(|instr_span| {
+                            after_next_left_pos < instr_span.left
+                                && instr_span.right < after_next_right_pos
+                        }) {
                             let error = ParserError::fold_has_instruction_after_next(*fold_span);
                             add_to_errors(&mut self.errors, *fold_span, Token::Next, error);
                         }
--- a/crates/air-lib/air-parser/src/parser/validator.rs
+++ b/crates/air-lib/air-parser/src/parser/validator.rs
@@ -421,20 +421,12 @@ impl<'i> VariableValidator<'i> {
             .iter()
             .filter(|(_, fold_iter_span)| xor_span.contains_span(*fold_iter_span))
         {
-            // if left_span.left <= fold_iter_span.left && fold_iter_span.right <= left_span.right {
             if left_span.contains_or_equal_span(*fold_iter_span) {
-                self.other_instrs_spans.retain(|instr_span| {
-                    !right_span.contains_or_equal_span(*instr_span) //.left <= instr_span.left && instr_span.right <= right_span.right)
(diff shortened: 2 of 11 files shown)


crates/air-lib/air-parser/src/parser/span.rs
```diff
@@ -39,6 +39,12 @@ impl Span {
     pub fn contains_span(&self, span: Self) -> bool {
         self.contains_position(span.left) && self.contains_position(span.right)
     }
+
+    // This presumes that a `span` that begins in `self` can not overlap
+    // and at the same time can not get outside the `self`.
+    pub fn contains_or_equal_span(&self, span: Self) -> bool {
+        self.left <= span.left && span.right <= self.right
```
monoid, line 46: Hard to read because both values are 4 letters and start with `s`.

crates/air-lib/air-parser/src/parser/validator.rs
```diff
@@ -517,6 +582,39 @@ impl<'i> ValidatorErrorBuilder<'i> {
...
+                    .filter(|fold_span| fold_span.contains_or_equal_span(*iterator_span))
+                {
+                    {
+                        let after_next_right_pos = fold_span.right;
+                        if self
+                            .validator
+                            .other_instrs_spans
+                            .iter()
+                            .find(|instr_span| {
+                                after_next_left_pos < instr_span.left
```
monoid, line 603: It should be clarified what these comparisons mean.

crates/air-lib/air-parser/src/parser/span.rs
raftedproc: Renamed

crates/air-lib/air-parser/src/parser/validator.rs
```diff
@@ -517,6 +582,33 @@ impl<'i> ValidatorErrorBuilder<'i> {
...
+                            after_next_left_pos < instr_span.left
+                                && instr_span.right < after_next_right_pos
+                        }) {
+                            let error = ParserError::fold_has_instruction_after_next(*fold_span);
+                            add_to_errors(&mut self.errors, *fold_span, Token::Next, error);
+                        }
+                    }
+                }
+            }
+        }
```
mikevoronov, line 608:
```rust
        for (iterator_name, iterator_span) in &self.validator.fold_iter_vec {
            let fold_spans = match self.validator.fold_vec.get_vec(iterator_name) {
                Some(fold_spans) => fold_spans,
                None => continue,
            };

            let fold_spans = fold_spans
                .iter()
                .filter(|fold_span| fold_span.contains_or_equal_span(*iterator_span));

            for fold_span in fold_spans {
                let after_next_span = Span::new(iterator_span.right, fold_span.right);

                let instructions_after_next =
                    self.validator.other_instrs_spans.iter().any(|instr_span| {
                        instr_span.contains_span(after_next_span)
                    });

                if instructions_after_next {
                    let error = ParserError::fold_has_instruction_after_next(*fold_span);
                    add_to_errors(&mut self.errors, *fold_span, Token::Next, error);
                }
            }
        }
```

crates/air-lib/air-parser/src/parser/validator.rs
```diff
@@ -383,6 +412,42 @@ impl<'i> VariableValidator<'i> {
...
+            debug_assert!(!self.other_instrs_spans.is_empty());
+            if let Some(last_instr_span) = self.other_instrs_spans.last() {
+                self.fold_vec.insert(
+                    iterator_name,
+                    Span::new(fold_span.left, last_instr_span.left),
+                );
+            }
+        } else {
+            self.fold_vec.insert(iterator_name, fold_span);
+        }
```
mikevoronov, line 449:
```rust
        if !fold_has_last {
            self.fold_vec.insert(iterator_name, fold_span);
            return;
        }

        debug_assert!(!self.other_instrs_spans.is_empty());
        if let Some(last_instr_span) = self.other_instrs_spans.last() {
            self.fold_vec.insert(
                iterator_name,
                Span::new(fold_span.left, last_instr_span.left),
            );
        }
```

crates/air-lib/air-parser/src/parser/validator.rs
```diff
@@ -383,6 +412,42 @@ impl<'i> VariableValidator<'i> {
...
+                    .retain(|instr_span| !left_span.contains_or_equal_span(*instr_span))
+            }
+        }
+    }
+
+    pub fn met_instr_w_span(&mut self, span: Span) {
+        self.other_instrs_spans.push(span);
+    }
+
+    fn met_fold(&mut self, iterator_name: &'i str, fold_has_last: bool, fold_span: Span) {
```
mikevoronov, line 438: Suggest renaming it into `met_fold_stream_or_map`

crates/air-lib/air-parser/src/parser/validator.rs
```diff
@@ -383,6 +412,42 @@ impl<'i> VariableValidator<'i> {
...
+        if fold_has_last {
+            debug_assert!(!self.other_instrs_spans.is_empty());
+            if let Some(last_instr_span) = self.other_instrs_spans.last() {
```
mikevoronov, line 441: I think that it's better to pass the last instruction and get its right span.

crates/air-lib/air-parser/src/parser/validator.rs
```diff
@@ -383,6 +412,42 @@ impl<'i> VariableValidator<'i> {
     fn met_iterator_definition(&mut self, iterator: &Scalar<'i>, span: Span) {
         self.met_iterator_definitions.insert(iterator.name, span);
     }
+
+    // Meeting xor the validator has to choose one of the two xor branches
+    // and clean up all instructions inserted into other_instr so far.
+    pub fn met_xor(&mut self, xor_span: Span, left_span: Span, right_span: Span) {
```
mikevoronov, line 418: I think that instead of deleting instructions, it's better to think in terms of allowed and prohibited spans. And in xor here you can just add values into an allowed span and then check them in `check_after_next_instr`. With that you can avoid traversing folds and all met instructions.

crates/air-lib/air-parser/src/parser/validator.rs
```diff
@@ -517,6 +582,33 @@ impl<'i> ValidatorErrorBuilder<'i> {
         self
     }
 
+    // The check loops over all iterators set with `next iterator` then searches
```
mikevoronov, line 585: It's not followed our comment coding style.

PR #761, opened 2023-11-29
Update benchmark results

gurinderu:


mikevoronov (approved):
utACK

monoid:
These benchmarks should be updated, especially after #780 is merged.

PR #762, opened 2023-11-29
chore(benchmark): restore parser-10000-100 bench

monoid:
A benchmark with empty datas that has long AIR code with lot of variable (re-)assignments.


raftedproc (approved):
utACK

PR #766, opened 2023-12-11
feat(air-parser)!: optimize `Instruction` type layout

monoid:
Instead of being boxed for each node, the `Instruction` type contains a boxed variable-size elements.  Thus `Instruction` is quite lean, and the allocator deals with variable-sized elements.

Total number of allocations is more or less same, but less space is wasted for unused memory: previously the Instruction's size was 112 bytes in WASM, now it is 16.

It reduces memory consumption on large AIR scripts (heap size decreased from 7.7Mb to 4.9Mb in parser-10000-100, for example).

This is a breaking change as the API changes (though the code that navigates the parsed tree generally should work as is).

### DIFF
--- a/air/src/lib.rs
+++ b/air/src/lib.rs
@@ -70,7 +70,7 @@ pub mod parser {
     pub use air_parser::ast::Instruction;
 
     /// Parse an AIR script to AST.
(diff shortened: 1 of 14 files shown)


monoid:
Wrong ticket number.

PR #767, opened 2023-12-11
feat(air-parser)!: optimize Instruction type layout

monoid:
Instead of being boxed for each node, the `Instruction` type contains a boxed variable-size elements.  Thus `Instruction` is quite lean, and the allocator deals with variable-sized elements.

Total number of allocations is more or less same, but less space is wasted for unused memory: previously the Instruction's size was 112 bytes in WASM, now it is 16.

It reduces memory consumption on large AIR scripts (heap size decreased from 7.7MiB to 4.625MiB in parser-10000-100, and for the new parser-calls-10000-100 benchmark, it decreased from 5.115MiB to 4.375MiB).

This is a breaking change as the API changes (though the code that navigates the parsed tree generally should work as is).

### DIFF
--- a/air/src/lib.rs
+++ b/air/src/lib.rs
@@ -70,7 +70,7 @@ pub mod parser {
     pub use air_parser::ast::Instruction;
 
     /// Parse an AIR script to AST.
-    pub fn parse(script: &str) -> Result<Box<Instruction<'_>>, String> {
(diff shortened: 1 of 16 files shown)


mikevoronov (approved):
utACK

raftedproc (approved):
utACK

PR #772, opened 2023-12-15
chore(deps): update near-sdk

mikevoronov:


monoid (approved):
utACK

tools/cli/air/Cargo.toml
```diff
@@ -32,7 +32,7 @@ bs58 = "0.5.0"
 zeroize = "1.6.0"
 
 # near
-near-sdk = { version = "4.1.1", optional = true }
+near-sdk = { git = "https://github.com/near/near-sdk-rs", rev = "0022685", optional = true }
```
monoid, line 35: It might be useful to mention why git is used and/or when we should switch back.

monoid (approved):
utACK

PR #774, opened 2023-12-19
feat(air-parser): VM-347 After next validator fold check

raftedproc:
This patch adds a new validator check that fails parsing if there is an instruction semantically following `next` in a `fold`.


air/tests/test_module/features/data_merging/data_merge.rs
```diff
@@ -219,6 +219,7 @@ fn stream_merge() {
 }
 
 #[test]
+#[ignore]
```
raftedproc, line 222: I have to restore these two tests.

crates/air-lib/air-parser/src/parser/validator.rs
```diff
@@ -30,6 +30,194 @@ use multimap::MultiMap;
...
+///    (call  ...)  <- instruction after next
+///  )
+/// )
+/// Note that the fold over scalar doesn't have this restriction.
+#[derive(Clone, Debug)]
+struct AfterNextCheckMachine<'name> {
+    /// stack for the machine.
+    stack: Vec<(CheckInstructionKind<'name>, Span)>,
+
+    /// lalrpop effectively parses from right to left thus `next`
```
monoid, line 60: Isn't it from inner to outer?

crates/air-lib/air-parser/src/parser/validator.rs
```diff
@@ -30,6 +30,194 @@ use multimap::MultiMap;
...
+    /// happens before fold. It is impossible to tell whether `next`
+    /// belongs to a scalar fold or not.
+    potentially_impossible_spans: Vec<(&'name str, Span)>,
+
+    /// This vector contains all spans where an instruction after next was met.
+    impossible_spans: Vec<Span>,
```
monoid, line 66: `denied_spans`, `violating_spans`, etc.  "Impossible" is too strong word.

crates/air-lib/air-parser/src/parser/validator.rs
```diff
@@ -30,6 +30,194 @@ use multimap::MultiMap;
...
+            stack: Default::default(),
+            potentially_impossible_spans: Default::default(),
+            impossible_spans: Default::default(),
+            is_enabled: true,
+        }
+    }
+}
+
+impl<'name> AfterNextCheckMachine<'name> {
+    fn impossible_spans_mut_iter(&mut self) -> std::slice::IterMut<Span> {
```
monoid, line 85: Is mut really needed?
And it seems all you need is an accessor to `&[Span]`.

crates/air-lib/air-parser/src/parser/validator.rs
```diff
@@ -30,6 +30,194 @@ use multimap::MultiMap;
...
+/// Note that the fold over scalar doesn't have this restriction.
+#[derive(Clone, Debug)]
+struct AfterNextCheckMachine<'name> {
+    /// stack for the machine.
+    stack: Vec<(CheckInstructionKind<'name>, Span)>,
+
+    /// lalrpop effectively parses from right to left thus `next`
+    /// happens before fold. It is impossible to tell whether `next`
+    /// belongs to a scalar fold or not.
+    potentially_impossible_spans: Vec<(&'name str, Span)>,
```
monoid, line 63: `suspectable_spans`/`unprocessed_spans`, etc.

crates/air-lib/air-parser/src/parser/validator.rs
```diff
@@ -30,6 +30,194 @@ use multimap::MultiMap;
...
+            }
+        }
+    }
+
+    fn after_next_check(&mut self, iterable: &'name str) {
+        let error_span = self
+            .potentially_impossible_spans
+            .iter()
+            .find(|(kind_iterable, _)| iterable == *kind_iterable)
+            .map(|(_, span)| span);
```
monoid, line 193: Can't we speed up it with a hashmap?

crates/air-lib/air-parser/src/parser/validator.rs
```diff
@@ -30,6 +30,194 @@ use multimap::MultiMap;
...
+                    ) if left_iterable == right_iterable => {
+                        self.stack
+                            .push((CheckInstructionKind::PivotalNext(left_iterable), span));
+                    }
+                    (Some((CheckInstructionKind::PivotalNext(iterator), _)), Some((..))) => {
+                        // potential failure but need to check when fold pops up.
+                        self.stack.push((CheckInstructionKind::Merging, span));
+                        self.potentially_impossible_spans.push((iterator, span));
+                    }
+                    (Some((..)), Some((CheckInstructionKind::PivotalNext(iterable), _))) => {
```
monoid, line 152: Why do you use both `..` and `_`?  Generally, `..` is used for "ignore rest of fields", not "all fields".

crates/air-lib/air-parser/src/parser/validator.rs
```diff
@@ -30,6 +30,194 @@ use multimap::MultiMap;
...
+impl<'name> AfterNextCheckMachine<'name> {
+    fn impossible_spans_mut_iter(&mut self) -> std::slice::IterMut<Span> {
+        self.impossible_spans.iter_mut()
+    }
+
+    fn met_instruction_kind(&mut self, instr_kind: CheckInstructionKind<'name>, span: Span) {
+        if !self.is_enabled {
+            return;
+        }
+        match instr_kind {
```
monoid, line 93: `use CheckInstructionKind::*;` will allow you to make the code shorter.

crates/air-lib/air-parser/src/parser/validator.rs
```diff
@@ -30,6 +30,194 @@ use multimap::MultiMap;
...
+                        // disable machine if Xoring invariant, namely there must be 2 kinds on a stack, is broken.
+                        self.stack.clear();
+                        self.is_enabled = false;
+                    }
+                }
+            }
+            CheckInstructionKind::Merging => {
+                let right_branch = self.stack.pop();
+                let left_branch = self.stack.pop();
+                match (left_branch, right_branch) {
```
monoid, line 139: It seems you can simplify the matches with `Option::zip`.

crates/air-lib/air-parser/src/parser/validator.rs
```diff
@@ -30,6 +30,194 @@ use multimap::MultiMap;
...
+            }
+            CheckInstructionKind::PopStack2 => {
+                self.stack.pop();
+                self.stack.pop();
+                self.stack.push((instr_kind, span))
+            }
+        }
+    }
+
+    fn after_next_check(&mut self, iterable: &'name str) {
```
monoid, line 188: Has same name as a field, makes it less convenient read.

crates/air-lib/air-parser/src/parser/validator.rs
```diff
@@ -30,6 +30,194 @@ use multimap::MultiMap;
...
+        iterator_name: &'name str,
+        span: Span,
+        instr_kind: &CheckInstructionKind<'name>,
+    ) {
+        let child = self.stack.pop();
+        match child {
+            Some((CheckInstructionKind::PivotalNext(checking_iterable), _)) => {
+                self.after_next_check(iterator_name);
+                self.stack
+                    .push((CheckInstructionKind::PivotalNext(checking_iterable), span));
```
monoid, line 210: You are recreating the match. It can be made more explicit if you capture the match and then use it.

crates/air-lib/air-parser/src/parser/validator.rs
```diff
@@ -30,6 +30,194 @@ use multimap::MultiMap;
...
+                let left_branch = self.stack.pop();
+                match (left_branch, right_branch) {
+                    (
+                        Some((CheckInstructionKind::PivotalNext(..), _)),
+                        Some((CheckInstructionKind::PivotalNext(..), _)),
+                    ) => {
+                        // `xor` has `next` in both branches. It is impossible
+                        // to tell which branch should poped up.
+                        self.stack.clear();
+                        self.is_enabled = false;
```
monoid, line 118: Let's introduce `self.disable()` given the statements above are repeated several times.

crates/air-lib/air-parser/src/parser/validator.rs
monoid: And I think branches are complex enough to be split into separate functions.

crates/air-lib/air-parser/src/parser/validator.rs
raftedproc: Renamed

crates/air-lib/air-parser/src/parser/validator.rs
raftedproc: Fixed

crates/air-lib/air-parser/src/parser/validator.rs
raftedproc: Done

crates/air-lib/air-parser/src/parser/validator.rs
raftedproc: Done

crates/air-lib/air-parser/src/parser/validator.rs
raftedproc: Fixed

crates/air-lib/air-parser/src/parser/validator.rs
raftedproc: Renamed

crates/air-lib/air-parser/src/parser/validator.rs
raftedproc: This forces me to use pattern matching with match guard uses matches! and this doesn't look any better than re-creating a capture.

crates/air-lib/air-parser/src/parser/validator.rs
raftedproc: This would add O(C) inserting potentially malformed span but it might worth it.

crates/air-lib/air-parser/src/parser/validator.rs
monoid: I mean something like this
```
Some((blabla@CheckInstructionKind::PivotalNext(_), _)) => {
  ...
  self.stack.push(blabla);
}
```

crates/air-lib/air-parser/src/parser/validator.rs
monoid: Child-parent may be better?  Can you elaborate on your objections?

raftedproc: I changed the wording.

mikevoronov (approved):
utACK

crates/air-lib/air-parser/src/parser/validator.rs
raftedproc: Fixed

crates/air-lib/air-parser/src/parser/validator.rs
```diff
@@ -30,6 +30,208 @@ use multimap::MultiMap;
...
+            Some(((pattern_kind @ PivotalNext(_), ..), ..))
+            | Some((_, (pattern_kind @ PivotalNext(_), _))) => {
+                // potential failure but need to check when fold pops up.
+                self.stack.push((pattern_kind, span));
+            }
+            Some(_) => {
+                self.stack.push((Xoring, span));
+            }
+            _ => {
+                // disable machine if Xoring invariant, namely there must be 2 kinds on a stack, is broken.
```
monoid, line 168: Can the invariant broken, but there are still 2 kinds on a stack, one is from some unrelated statement?

Does this invariant relates to correctness of parsing (i.e. it can break only if parsing/parser is incorrect)?

raftedproc: No.
Yes. Changed semantics might also result in a failure.

monoid (approved):
utACK

PR #775, opened 2023-12-20
feat(data)!: values are binary blobs

monoid:
This is a preparation PR for RKYV transition.


crates/air-lib/interpreter-data/src/value.rs
```diff
@@ -0,0 +1,69 @@
...
+use crate::JValue;
+
+use serde::{Deserialize, Serialize};
+
+use std::cell::RefCell;
+use std::rc::Rc;
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(transparent)]
+pub struct RawValue {
```
mikevoronov, line 26: Would it be better to implement it as enum with two variants: Parsed and Unparsed?

crates/air-lib/interpreter-data/src/value.rs
```diff
@@ -0,0 +1,69 @@
...
+    raw: Box<str>,
+
+    #[serde(skip)]
+    parsed: RefCell<Option<Rc<JValue>>>,
```
mikevoronov, line 30: looks like `Cell` is enough here

crates/air-lib/interpreter-data/src/value.rs
monoid: That would require manual implementation of serde type, unfortunately.
One might use this type for `parsed` field, but it would be same as `Option` (though it is not bad).

crates/air-lib/interpreter-data/src/value.rs
monoid: It seems one cannot clone `Cell` contents directly, at least with stable API.

crates/air-lib/interpreter-data/src/value.rs
```diff
@@ -0,0 +1,69 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+use crate::JValue;
+
+use serde::{Deserialize, Serialize};
```
mikevoronov, line 19: let's unwrap them

crates/air-lib/interpreter-data/src/value.rs
mikevoronov: I think that serde implementation would be OK, but we can do it in separate PR.

mikevoronov (approved):
utACK

crates/air-lib/interpreter-data/src/value.rs
monoid: Something like `serde_with` might help to make it simple.
But I not sure that it worth it given `RawValue` is a type with interior mutability.  That mutable part is placed in the separate field.

mikevoronov: I think we can we can do smth like that:

```rust
struct RawValue {
  impl: RefCell<RawValueImpl>
}
```

I'd say that it'll make code more readable.

mikevoronov: Moreover it'll require less memory

monoid: I think it is 4 words either way.

PR #776, opened 2023-12-20
feat(avm-client): implement pluggable formats in JS

monoid:
For parsing and producing call requests and call results in AquaVM-compatible way with JSON and MessagePack.

Multicodec representation is also supported, both JSON and MessagePack can be used as input.

### DIFF
--- a/avm/client/src/avmHelpers.ts
+++ b/avm/client/src/avmHelpers.ts
@@ -15,9 +15,19 @@
  */
 
 import { CallResultsArray, InterpreterResult, CallRequest, RunParameters, JSONArray, JSONObject } from './types';
+import { JsonRepr } from './formats'
 
 const decoder = new TextDecoder();
 const encoder = new TextEncoder();
+
+// Have to match the air-interpreter-interface.
+const callRequestsRepr = new JsonRepr();
+// Have to match the air-interpreter-interface.
+const argumentRepr = new JsonRepr();
+// Have to match the air-interpreter-interface.
+const tetrapletRepr = new JsonRepr();
+
+const callResultsRepr = new JsonRepr();
 
 /**
  * Encodes arguments into JSON array suitable for marine-js
@@ -44,7 +54,7 @@ export function serializeAvmArgs(
         };
     }
 
-    const encoded = encoder.encode(JSON.stringify(callResultsToPass));
+    const encodedCallResults = callResultsRepr.toBinary(callResultsToPass)
     const runParamsSnakeCase = {
         init_peer_id: runParams.initPeerId,
         current_peer_id: runParams.currentPeerId,
@@ -55,7 +65,7 @@ export function serializeAvmArgs(
         particle_id: runParams.particleId,
     };
 
-    return [air, Array.from(prevData), Array.from(data), runParamsSnakeCase, Array.from(encoded)];
+    return [air, Array.from(prevData), Array.from(data), runParamsSnakeCase, Array.from(encodedCallResults)];
 }
 
 /**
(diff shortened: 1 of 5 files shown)


avm/client/src/formats.ts
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as msgpackr from "msgpackr"
+import * as multicodec from "multicodec"
```
akim-bow, line 18:
```suggestion
import msgpackr from "msgpackr"
import multicodec from "multicodec"
```
TS does this for you

avm/client/src/formats.ts
```diff
@@ -0,0 +1,99 @@
...
+export class MulticodecRepr implements Representation {
+    serializer: Representation & Multiformatable
+
+    constructor(serializer: Representation & Multiformatable) {
+        this.serializer = serializer
+    }
+
+    fromBinary(data: Uint8Array): object {
+        let code = multicodec.getCodeFromData(data)
+        var repr = null;
```
akim-bow, line 79: Var is kinda outdated in new code

```suggestion
        let repr = null;
```

avm/client/src/formats.ts
```diff
@@ -0,0 +1,99 @@
...
+            repr = new JsonRepr()
+        } else if (code == multicodec.MESSAGEPACK) {
+            repr = new MsgPackRepr()
+        }
+
+        if (repr === null) {
+            throw "Unknown code " + code + "in multiformat data " + data
+        } else {
+            return repr.fromBinary(multicodec.rmPrefix(data))
+        }
```
akim-bow, line 91:
```suggestion
        if (repr === null) {
            throw "Unknown code " + code + "in multiformat data " + data
        }
        
        return repr.fromBinary(multicodec.rmPrefix(data))
```

No need to do excessive branches

avm/client/src/formats.ts
monoid: So, let-variables are mutable too?

monoid: And `repr: Representation` can contain `null`, can't it?

avm/client/src/avmHelpers.ts
```diff
@@ -65,15 +75,15 @@ export function serializeAvmArgs(
  */
 export function deserializeAvmResult(result: any): InterpreterResult {
```
akim-bow, line 76: What about changing `any` type here to specific value? I think it would be much easier for consumers to typecheck these calls.

avm/client/src/avmHelpers.ts
```diff
@@ -83,15 +93,13 @@ export function deserializeAvmResult(result: any): InterpreterResult {
...
             throw "Couldn't parse arguments: " + e + '. Original data is: ' + callRequest.arguments;
         }
 
         try {
-            const tetrapletsStr = decoder.decode(new Uint8Array(callRequest.tetraplets));
-            tetraplets = JSON.parse(tetrapletsStr);
+            tetraplets = tetrapletRepr.fromBinary(callRequest.tetraplets)
         } catch (e) {
             throw "Couldn't parse tetraplets: " + e + '. Original data is: ' + callRequest.tetraplets;
         }
```
akim-bow, line 105: below you have `key as any`
Probably you need to rewrite it as Number(key).
`for ... in ` cycle gives only string key and judging by the output type you'd like to have numbers

avm/client/src/formats.ts
akim-bow: Yes, `let` variables are mutable, `const` - not mutable.

akim-bow: Changed the suggestion to include null in union types. By default null is not included

akim-bow: After this change looks like you can safely remove `"strictNullChecks": false`.

Also i would suggest to remove `"noImplicitAny": false` but it will require some refactoring and i think we can do this iteratively in the future PRs.

avm/client/src/avmHelpers.ts
monoid: It worth changing it in a separate PR.

avm/client/src/avmHelpers.ts
monoid: Same for that separate PR.

avm/client/src/avmHelpers.ts
```diff
@@ -15,10 +15,20 @@
...
 const encoder = new TextEncoder();
 
+// Have to match the air-interpreter-interface.
+const callRequestsRepr = new JsonRepr();
+// Have to match the air-interpreter-interface.
+const argumentRepr = new JsonRepr();
+// Have to match the air-interpreter-interface.
+const tetrapletRepr = new JsonRepr();
+
+const callResultsRepr = new JsonRepr();
```
monoid, line 30: Same comment about air-interpreter-interface should be added here.

akim-bow:
Looks good. Please remeber to create a task to fix at least some of the `any` values as they are disabling typescript checking in consumer's code.

akim-bow (approved):

PR #779, opened 2023-12-25
chore(perf): database data update

raftedproc:
This is to update the timings in bench database after the recent AquaVM release.


mikevoronov (approved):
utACK

monoid:
I suggest to wait till I enable MsgPack for calls, and rerun. It will be more sensible.

mikevoronov:
> I suggest to wait till I enable MsgPack for calls, and rerun. It will be more sensible.

It's needed only to test [VM-377](https://github.com/fluencelabs/aquavm/pull/774), so it's OK to merge it whenever we're close to merge VM-377.

PR #780, opened 2023-12-25
feat(interface)!: Use MessagePack for calls

monoid:
Top-level structs use multicodec-prefixed MessagePack, the nested structures (arguments, tetraplets) use just ordinary MessagePack.

The old benchmark was against MessagePack version too, so the numbers haven't changed much.


air/src/preparation_step/preparation.rs
```diff
@@ -152,6 +159,13 @@ fn make_exec_ctx(
     Ok(ctx)
 }
 
+fn unjs_call_results(call_results: CallResults) -> Result<HashMap<u32, CallServiceResult>, ParseIntError> {
+    call_results
+        .into_iter()
+        .map(|(id, res)| id.parse().map(|id| (id, res)))
```
raftedproc, line 165: It is hard to read this line b/c both variable names are the same in both map() calls.

air/src/preparation_step/preparation.rs
```diff
@@ -152,6 +159,13 @@ fn make_exec_ctx(
     Ok(ctx)
 }
 
+fn unjs_call_results(call_results: CallResults) -> Result<HashMap<u32, CallServiceResult>, ParseIntError> {
```
raftedproc, line 162: Could you ellaborate on the meaning of `unjs` prefix?

air/src/execution_step/execution_context/context.rs
```diff
@@ -75,7 +76,7 @@ pub(crate) struct ExecutionCtx<'i> {
     pub(crate) last_call_request_id: u32,
 
     /// Contains all executed results from a host side.
-    pub(crate) call_results: CallResults,
+    pub(crate) call_results: HashMap<u32, CallServiceResult>,
```
raftedproc, line 79: It is unclear what is u32 key is for just looking at the struct so it is worth to elaborate on this in the comments.

raftedproc: I would name `HashMap<u32, CallServiceResult>` type.

air/src/preparation_step/errors.rs
```diff
@@ -98,6 +100,9 @@ pub enum PreparationError {
     /// Failed to check peers' signatures.
     #[error(transparent)]
     DataSignatureCheckError(#[from] DataVerifierError),
+
+    #[error("invalid call result ID: {0}")]
+    InvalidCallResultId(ParseIntError),
```
raftedproc, line 105: Could you elaborate in which way the call result is invalid? Its call-id is unexpected, e.g. too high ?

benches/PERFORMANCE.txt
```diff
@@ -752,688 +752,688 @@ Machine c1f3ea5950db0a10b44da931c25774d64ab25084f47d504f72f311e694550ff1:
...
+            populate_outcome_from_contexts: 255.00µs
               CallRequestsRepr.serialize: 13.00µs
-              InterpreterData::serialize: 44.00µs
+              InterpreterData::serialize: 40.00µs
         signing_step::sign_produced_cids: 158.00µs
         verification_step::verify: 35.00µs
           verify: 10.00µs
-    parser-calls-10000-100 (25.49ms; 54.375 MiB, 54.375 MiB): multiple calls parser benchmark
-      air::runner::execute_air: 25.49ms
+    parser-10000-100 (30.60ms; 54.625 MiB, 54.625 MiB): long air script with lot of variable assignments
```
raftedproc, line 1374: +5 secs. Is it an expected degradation?

air/src/preparation_step/preparation.rs
monoid: That's same value represented differently.  I think it is OK for Rust.

air/src/execution_step/execution_context/context.rs
monoid: Naming makes sense.

air/src/preparation_step/errors.rs
monoid: Call IDs too high do not lead to `ParseIntError`.  But I should check if the message is clear to a user.

air/src/execution_step/execution_context/context.rs
mikevoronov: Actually, I don't understand why it's a concern that one need to go to the definition of the type to see the key type? One can replace `CallResults` with explicit type, but it's used in many places in the code base, so there will be a lot of changes and the explicit type is much more verbose.

air/src/preparation_step/preparation.rs
```diff
@@ -134,14 +139,16 @@ fn make_exec_ctx(
     signature_store: SignatureStore,
     run_parameters: &RunParameters,
 ) -> PreparationResult<ExecutionCtx<'static>> {
-    let call_results = measure!(
+    let call_results_js = measure!(
```
mikevoronov, line 142: what is the meaning of the `js`suffix here?

air/src/preparation_step/preparation.rs
```diff
@@ -134,14 +139,16 @@ fn make_exec_ctx(
...
         CallResultsRepr
             .deserialize(call_results)
             .map_err(|e| PreparationError::call_results_de_failed(call_results.clone(), e))?,
         tracing::Level::INFO,
         "CallResultsRepr.deserialize",
     );
 
+    let call_results = unjs_call_results(call_results_js).map_err(PreparationError::InvalidCallResultId)?;
```
mikevoronov, line 150: I think that one day we can abstract from `u32` in call results, we can also use Strings ot Vec<u8> for them if it's more convenient b/c for the host side it's just a tag which internal structure is known only to AquaVM.

crates/air-lib/interpreter-interface/src/call_service_result.rs
```diff
@@ -16,17 +16,20 @@
...
-use air_interpreter_sede::JsonFormat;
+use air_interpreter_sede::MsgPackMultiformat;
 use air_interpreter_sede::Representation;
 use serde::Deserialize;
 use serde::Serialize;
 use serde_json::Value as JValue;
 use std::collections::HashMap;
 
-pub type CallResults = HashMap<u32, CallServiceResult>;
+/// This is a map from String to service result for compatiblity with JavaScript.
```
mikevoronov, line 26:
```suggestion
/// This is a map from a String to a service result for compatibility with JavaScript.
```

crates/air-lib/interpreter-interface/src/call_service_result.rs
```diff
@@ -16,17 +16,20 @@
...
+/// Binary format implementations like `rmp-serde` do not bother converting keys from strings, unlike `serde_json`.
+/// So, we do it manually for all formats
```
mikevoronov, line 28:
```suggestion
/// So, we do it manually for all formats.
```

air/src/preparation_step/preparation.rs
monoid: JS-compatible representation of call results.

air/src/preparation_step/preparation.rs
monoid: Well, we may query call ID by its string representation, I thought about it.

air/src/execution_step/execution_context/context.rs
monoid: @mikevoronov I don't quite understand what is your point.  `CallResult` is an interface type that now has `String` keys.

air/src/preparation_step/preparation.rs
monoid: Switched to strings for call IDs.

air/src/preparation_step/preparation.rs
monoid: The code was removed.

crates/air-lib/interpreter-interface/src/call_service_result.rs
```diff
@@ -16,17 +16,20 @@
...
+use air_interpreter_sede::MsgPackMultiformat;
 use air_interpreter_sede::Representation;
 use serde::Deserialize;
 use serde::Serialize;
 use serde_json::Value as JValue;
 use std::collections::HashMap;
 
-pub type CallResults = HashMap<u32, CallServiceResult>;
+/// This is a map from a String to a service result for compatibility with JavaScript.
+/// Binary format implementations like `rmp-serde` do not bother converting keys from strings, unlike `serde_json`.
```
mikevoronov, line 27: looks like it's obsolete

mikevoronov (approved):
utACK

crates/air-lib/interpreter-interface/src/call_service_result.rs
monoid: No, it explains why `String` is used at the next line.

PR #790, opened 2024-01-09
chore(benches): this is a set of RAM intensive benchmarks

raftedproc:
To understand memory consumption patterns in AquaVM one has to run these memory-intensive benchmarking tests.


junk/gen-bench-data/src/main.rs
```diff
@@ -735,6 +751,257 @@ fn big_values_data() -> Data {
...
 
+fn generate_random_data(random_data_size: usize) -> Vec<u8> {
+    use rand::thread_rng;
+    use rand::Rng;
+
+    // hex::encode later prints out every byte as 2 bytes sequence.
+    let random_data_size = random_data_size / 2 * MB;
+
+    let mut random_data = Vec::<u8>::with_capacity(random_data_size);
+    let mut rng = thread_rng();
```
monoid, line 762: It makes the benches unreproducible (even if no variability is expected).  I suggest using a fixed seed.

junk/gen-bench-data/src/main.rs
```diff
@@ -735,6 +751,257 @@ fn big_values_data() -> Data {
...
+
+    while random_data.len() < random_data_size {
```
monoid, line 764: What is the point of such loop?  Can't you just use `rng.fill(&mut random_data[..]);`, filling `random_data` with zeroes, which is cheap?

```
let random_data = vec![0u8; random_data_size];
rng.fill(&mut random_data[..]);
```

That's it.

junk/gen-bench-data/src/main.rs
```diff
@@ -735,6 +751,257 @@ fn big_values_data() -> Data {
...
+        let remaining_space = random_data_size - random_data.len();
+        let chunk_size = std::cmp::min(rng.gen_range(1..=remaining_space), remaining_space);
```
monoid, line 766: Is this `min` really needed?

monoid (approved):
utACK

PR #796, opened 2024-01-15
chore(performance): micro tweak to reduce peak consumption

raftedproc:
This reduces memory footprint for benchmarks with huge input data.


monoid (approved):
utACK

PR #799, opened 2024-01-16
feat(performance): reduced stack size

raftedproc:
This patch reduces overall AquaVM runtime memory consumption decreasing WASM VM stack down to 8 MB that is common value for stack in Linux distros [fixes VM-408]


mikevoronov (approved):
utACK

PR #800, opened 2024-01-17
chore(performance,tests):  hybrid workload memory intensive test

raftedproc:
This adds a mixed workload test to monitor memory consumption.


monoid:
It should be `chore(benches)`.

mikevoronov (approved):
utACK

PR #801, opened 2024-01-17
fix(performance): data printout removal some preparation step errors increases data consumption printout data out

raftedproc:
Some preparation step errors increases data consumption printout data out. This patch removes data from preparation step errors to save some RAM bytes.

### DIFF
--- a/air/src/preparation_step/errors.rs
+++ b/air/src/preparation_step/errors.rs
@@ -37,8 +37,7 @@ pub enum PreparationError {
     /// Errors occurred on executed trace deserialization.
     #[error(
         "an error occurred while data deserialization: {error:?}.\n\
-        AquaVM version is {} and it expect data of {} version,\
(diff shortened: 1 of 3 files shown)


air/src/preparation_step/errors.rs
```diff
@@ -39,57 +38,41 @@ pub enum PreparationError {
     #[error(
         "an error occurred while data deserialization: {error:?}.\n\
         AquaVM version is {} and it expect data of {} version,\
-        it's failed to get version of AquaVM produced this data.\n\
-        Data: {raw_data:?}",
+        it's failed to get version of AquaVM produced this data.",
```
monoid, line 41: Now "this" looks strange, IMHO.

mikevoronov (approved):
I suggest renaming this PR title into
`fix(perf)!: avoid printing data in errors`

monoid (approved):
utACK

PR #805, opened 2024-01-22
feat!: use marine with wasmtime 0.13.0

ValeryAntopol:


ValeryAntopol:
Updated to a newer version in #768

PR #806, opened 2024-01-23
chore(performance): use `fluence-blake3` with wasm32 SIMD support

monoid:
Enable BLAKE3 SIMD support on the wasm32 target.

It gives a noticeable performance improvement.


mikevoronov (approved):
utACK

PR #808, opened 2024-01-25
feat(cli): use eyre in air-cli

monoid:
Instead of `anyhow`, use `eyre` to reduce number of deps.  The `eyre` crate is already used in `avm-server` and `fluence-keypair`.


mikevoronov (approved):
utACK

PR #813, opened 2024-02-05
feat(execution-engine): Rc-based JSON value

monoid:
Add `air-interpreter-value` crate that contains `JValue` type that describes a JSON value, but is Rc-based, and its subelements can be cheaply copied and use in other JSON values.

It reduces memory and CPU consumption for AquaVM.

Please note that the new type is used only in the AquaVM internals and in not visible on the interface level.


air/src/execution_step/lambda_applier/applier.rs
```diff
@@ -280,13 +277,14 @@ fn select_by_functor_from_scalar(value: &JValue, functor: &Functor) -> Execution
...
-            Ok(serde_json::json!(length))
+            Ok(length.into())
         }
     }
 }
 
-impl<'value> LambdaResult<'value> {
-    fn from_cow(result: Cow<'value, JValue>, tetraplet_idx: usize) -> Self {
+impl LambdaResult {
+    // TODO rename
```
monoid, line 286: TODO

mikevoronov (approved):
utACK

PR #818, opened 2024-02-19
feat(avm-server): add a convenience method to hide implementation from nox

raftedproc:
This convenience method hides internal repr from nox

### DIFF
--- a/crates/air-lib/interpreter-interface/src/interpreter_outcome.rs
+++ b/crates/air-lib/interpreter-interface/src/interpreter_outcome.rs
@@ -117,14 +117,15 @@ impl InterpreterOutcome {
             ));
         }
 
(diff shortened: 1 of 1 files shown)


crates/air-lib/interpreter-interface/src/interpreter_outcome.rs
```diff
@@ -74,6 +74,12 @@ impl SoftLimitsTriggering {
             call_result_size_limit_exceeded,
         }
     }
+
+    pub fn is_triggered(&self) -> bool {
```
mikevoronov, line 78: Probably it should be `is_limits_exceeded`

mikevoronov (approved):
utACK

PR #819, opened 2024-02-20
feat(performance)!: update perf database and trigger a minor version bump

raftedproc:
This patch bumps minor version for the delivered artifacts.


mikevoronov (approved):
utACK

PR #822, opened 2024-02-22
chore: Rust channel update dedicated to ENIAC Day

monoid:
It requires updating minor version of curve25519-dalek crate in the Cargo.lock.  If you need to use older version of nightly rustc, use curve25519-dalek-4.1.1.  It affects Cargo.lock only.


raftedproc (approved):
utACK

PR #824, opened 2024-02-23
fix(avm-client)!: this fixes the last argument type for WASM module 'invoke' function

raftedproc:
This fixes the 'invoke' call incorrect argument type issue for js-client.

### DIFF
--- a/avm/client/src/avmHelpers.ts
+++ b/avm/client/src/avmHelpers.ts
@@ -27,6 +27,7 @@ const tetrapletRepr = new MsgPackRepr();
 const callResultsRepr = new MulticodecRepr(new MsgPackRepr());
 //
 const defaultAquaVMRuntimeMemoryLimit = Number("4294967296");
+const aquaVMRuntimeHardLimitsAreDisabled = Boolean(false);
 
 /**
  * Encodes arguments into JSON array suitable for marine-js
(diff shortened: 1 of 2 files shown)


avm/client/src/avmHelpers.ts
```diff
@@ -66,7 +67,7 @@ export function serializeAvmArgs(
         air_size_limit: defaultAquaVMRuntimeMemoryLimit,
         particle_size_limit: defaultAquaVMRuntimeMemoryLimit,
         call_result_size_limit: defaultAquaVMRuntimeMemoryLimit,
-        hard_limit_enabled: defaultAquaVMRuntimeMemoryLimit,
+        hard_limit_enabled: aquaVMRuntimeHardLimitsAreDisabled,
```
monoid, line 70: Is it enabled or disabled?

avm/client/src/avmHelpers.ts
```diff
@@ -27,6 +27,7 @@ const tetrapletRepr = new MsgPackRepr();
 const callResultsRepr = new MulticodecRepr(new MsgPackRepr());
 //
 const defaultAquaVMRuntimeMemoryLimit = Number("4294967296");
+const aquaVMRuntimeHardLimitsAreDisabled = Boolean(false);
```
akim-bow, line 30: Why not just put values directly?
```suggestion
const defaultAquaVMRuntimeMemoryLimit = 4294967296;
const aquaVMRuntimeHardLimitsAreDisabled = false;
```

avm/client/src/avmHelpers.ts
raftedproc: disabled.

avm/client/src/avmHelpers.ts
raftedproc: Changed

monoid (approved):
utACK

PR #844, opened 2024-06-24
chore: change license to AGPL-3

mikevoronov:


crates/air-lib/interpreter-value/src/value/mod.rs
```diff
@@ -1,17 +1,20 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
 
 /*
```
monoid, line 20: I think now we have explicitly quote the license text below in this crate.
Previously it was same license as above, now it is not the case.

FluenceCLA.md
```diff
@@ -0,0 +1,107 @@
+Thank you for your interest in Fluence Labs projects.
```
folex, line 1: seems like i have missed this. let's rename to cloudless labs.

monoid (approved):
utACK

PR #845, opened 2024-07-19
fix(execution-engine): quickfix for canon map iteration order

monoid:
A O(n) fix for the canon map reverse iteration bug.


raftedproc (approved):
untested ACK

InversionSpaces (approved):

PR #851, opened 2024-07-22
chore: Rust channel update dedicated to Panamerican Engineer's Day 

monoid:
It fixes compilation issues with serde that break crate publishing.


raftedproc (approved):
untested ACK

PR #853, opened 2024-07-31
feat(air): embedded Starlark script support

monoid:
1. Add a Starlark interpreter with redefined `fail(code, message)` function call, and `get_value(N)` and `get_tetraplet(N)` functions.
2. Add new `embed [args] (# starlark-script #) optional-output` instruction.


monoid:
The `embed` instruction should be documented.

